PR #6, opened 2020-08-20
EVA-2126 - Script to check if the mapping weight and multi copy actually matches

tcezard:


tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+                erroneous_recovered1 += 1
+            elif operation2 and operation2[0]['eventType'] == 'UPDATED':
+                erroneous_recovered2 += 1
+            else:
+                print(
+                    "rs%s Should have more than 1 variant because mapWeight = %s" % (
+                    str(variant['accession']), variant['mapWeight'])
+                )
+                print("Found %s operation for clustered variants" % len(operation1))
+                print("Found %s operation for submitted variants" % len(operation2))
```
jmmut, line 50: I think the next comment doesn't have any impact and we can go ahead, but let's not forget this:
Regarding counting SVE operations as multimap occurrences (https://github.com/EBIvariation/eva-accession/blob/8899037231cf48e698787c6b642fd552ad0044df/eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/ClusteringWriter.java#L224):
   
> Note that for submitted variants the test is not this simple, as 1:1000:A:T and 1:1000:A:G can be present in the
> same assembly and still not classify as multimap.

if 2 SNV submitted variants in the same position are linked to the same RS, and they are deprecated, there will be 2 operations but only 1 CVE. This won't affect this code snippet significantly but I think it may report a higher number of SVE ops.

jmmut (approved):

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+        dbsnp_cve_op_collection = accessioning_mongo_handle[database_name]["dbsnpClusteredVariantOperationEntity"]
+        dbsnp_sve_op_collection = accessioning_mongo_handle[database_name]["dbsnpSubmittedVariantOperationEntity"]
+
+        for variant in list_variant_erroneous:
+            operation1 = list(dbsnp_cve_op_collection.find({'accession': variant['accession']}))
+            operation2 = list(dbsnp_sve_op_collection.find({'inactiveObjects.rs': variant['accession']}))
+
+            if operation1 and operation1[0]['eventType'] == 'MERGED':
+                erroneous_recovered1 += 1
+            elif operation2 and operation2[0]['eventType'] == 'UPDATED':
```
andresfsilva, line 42: To have an `UPDATE` operation does not mean the event was registered exclusively because the submitted variant was declustered. There are other reasons for writing an update event. What we have done in the past to check declustering events is to look at the reason. It is not the safest because that is a text field but I think for the declustering it will be the same for all cases. You can see one validation for that [here](https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-release/src/main/java/uk/ac/ebi/eva/accession/release/batch/io/MergedVariantMongoReader.java#L153)

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+        for variant in cursor:
+            count = dbsnp_cve_collection.count({'accession': variant['accession']})
+            count_variants += 1
+            if count < 2:
+                list_variant_erroneous.append(variant)
+
+        dbsnp_cve_op_collection = accessioning_mongo_handle[database_name]["dbsnpClusteredVariantOperationEntity"]
+        dbsnp_sve_op_collection = accessioning_mongo_handle[database_name]["dbsnpSubmittedVariantOperationEntity"]
+
+        for variant in list_variant_erroneous:
```
andresfsilva, line 36: suggest to change `list_variant_erroneous` for something like `clustered_variants_wrong_map_weight` to easily know why they are erroneous

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+            operation1 = list(dbsnp_cve_op_collection.find({'accession': variant['accession']}))
```
andresfsilva, line 37: suggest to change `operation1` to `clustered_variant_operation`

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+            operation2 = list(dbsnp_sve_op_collection.find({'inactiveObjects.rs': variant['accession']}))
```
andresfsilva, line 38: suggest to change `operation2` to `submitted_variant_operation`

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+                erroneous_recovered2 += 1
+            else:
+                print(
+                    "rs%s Should have more than 1 variant because mapWeight = %s" % (
+                    str(variant['accession']), variant['mapWeight'])
+                )
+                print("Found %s operation for clustered variants" % len(operation1))
+                print("Found %s operation for submitted variants" % len(operation2))
+
+        print("Checked %s variants" % count_variants)
```
andresfsilva, line 52: Maybe it is important to specify when we are referring to submitted or clustered variants to avoid confusion

andresfsilva (changes requested):
The code looks good overall but left some comments that worth checking

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+    return pymongo.MongoClient(mongo_connection_uri)
+
+
+def check_mapping_weight(mongo_host, database_name, username, password, reference_accession):
+    """
+    Connect to mongodb and retrieve all variants the should be updated, Check their key and update them in bulk.
+    """
+    with get_mongo_connection_handle(mongo_host, username=username, password=password) as accessioning_mongo_handle:
+        dbsnp_cve_collection = accessioning_mongo_handle[database_name]["dbsnpClusteredVariantEntity"]
+        cursor = dbsnp_cve_collection.find({'mapWeight': {'$gt': 1}, 'asm': reference_accession})
```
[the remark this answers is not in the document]
tcezard: I changed it because the query makes more sense like that but it won't have any effect on the query performance.

tasks/eva_2126/check_mapping_weight.py
tcezard: I don't think they have the wrong mapping weight, most likely is they have fewer copy than expected because of previous operation. Actually before we test we do not know why they are erroneous.

tasks/eva_2126/check_mapping_weight.py
tcezard: Changed it.

tasks/eva_2126/check_mapping_weight.py
tcezard: Changed it to updated_sve_operation.

tasks/eva_2126/check_mapping_weight.py
tcezard: Changed it to merged_cve_operation

tasks/eva_2126/check_mapping_weight.py
tcezard: I renamed them inconsistent but did not define the inconsistency in the variable name.

tasks/eva_2126/check_mapping_weight.py
tcezard: Thank you. I changed it.

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+    mongo_connection_uri = "mongodb://"
+    if username and password:
+        mongo_connection_uri += '%s:%s@' % (username, quote_plus(password))
+    mongo_connection_uri += '%s:%s/%s' % (host, port, authentication_database)
+    return pymongo.MongoClient(mongo_connection_uri)
+
+
+def check_mapping_weight(mongo_host, database_name, username, password, reference_accession):
+    """
+    Connect to mongodb and retrieve all variants the should be updated, Check their key and update them in bulk.
```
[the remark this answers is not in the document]
tcezard: updated the comment

tasks/eva_2126/check_mapping_weight.py
```diff
@@ -0,0 +1,71 @@
...
+        print("Found %s erroneous variants that are due to declustering" % erroneous_recovered2)
+
+
+def main():
+    argparse = ArgumentParser()
+    argparse.add_argument('--host', help='', required=True)
+    argparse.add_argument('--database_name', help='', required=True)
+    argparse.add_argument('--username', help='', default=None)
+    argparse.add_argument('--password', help='', default=None)
+    argparse.add_argument('--reference_accession', help='', required=True)
```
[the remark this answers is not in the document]
tcezard: changed it.

tasks/eva_2126/check_mapping_weight.py
tcezard: Thanks for the comment. The number operation does not actually matter since I'm assuming that one operation is enough for the number of variant to be incorrect. The `print` was here to detect if there were any other operations present that might explain the missing variants

PR #7, opened 2020-08-20
EVA-2124 create a script to fix the contigs in the accessioning service

jmmut:


tasks/eva_2124/tests/test_correct_assembly_for_study.py
```diff
@@ -0,0 +1,130 @@
...
+        self.connection_handle[self.admin_db].command("dropUser", self.username)
+        self.connection_handle.close()
+
+    def test_correct(self):
+        fixed = correct(self.username, self.password, self.host, studies=['PRJEB30080'],
+                        assembly_accession='GCA_000181335.3')
+        self.assertEqual(fixed, 1)
+        variant = (self.connection_handle[self.eva_database][self.variant_collection].find_one({'seq': 'GCA_000181335.3'}))
+        self.assertEqual(variant['contig'], 'CM001381.2')
+        variant = (self.connection_handle[self.eva_database][self.variant_collection].find_one(
```
tcezard, line 98: I was going to say that the parenthesis around the `self.connection_handle` calls are superfluous  but then I checked my code from EVA-2068 and realised I did the same thing :sweat: :sweat:

tasks/eva_2124/correct_assembly_for_study.py
```diff
@@ -0,0 +1,112 @@
...
+        return contig, True
+
+    raise Exception('could not find synonym for contig {}'.format(contig))
+
+
+def correct(mongo_user, mongo_password, mongo_host, studies, assembly_accession, chunk_size = 1000):
+    """
+    Connect to mongodb and retrieve all variants the should be updated, check their key and update them in bulk.
+    """
+    with get_mongo_connection_handle(
```
tcezard, line 44: I have an issue with this function because It force us to have a username and password. This mean you won't be able to test on EMBASSY.  
I changed it slightly wand will post it back when PR https://github.com/EBIvariation/eva-common-pyutils/pull/1 is merged.
Here is what I have: 
```python
def get_mongo_connection_handle(host, port=27017, username=None, password=None, authentication_database="admin") -> pymongo.MongoClient:
    mongo_connection_uri = "mongodb://"
    if username and password:
        mongo_connection_uri += '%s:%s@' % (username, quote_plus(password))
    mongo_connection_uri += '%s:%s/%s' % (host, port, authentication_database)
    return pymongo.MongoClient(mongo_connection_uri)
```

tasks/eva_2124/correct_assembly_for_study.py
```diff
@@ -0,0 +1,112 @@
...
+    drop_statements.clear()
+    return total_inserted, total_dropped
+
+
+def main():
+    argparse = ArgumentParser()
+    argparse.add_argument('--mongo_user', help='user to connect to mongodb', required=True)
+    argparse.add_argument('--mongo_password', help='password to connect to mongodb', required=True)
+    argparse.add_argument('--mongo_host', help='host to connect to mongodb', required=True)
+    argparse.add_argument('--studies', help='The studies in the assembly to correct', required=True)
```
tcezard, line 103: You can allow space separated list of studies with the parameters `nargs='+'`. `args.studies` will then return a list.

tasks/eva_2124/correct_assembly_for_study.py
```diff
@@ -0,0 +1,112 @@
...
+    total_dropped += result_drop.deleted_count
+    insert_statements.clear()
+    drop_statements.clear()
+    return total_inserted, total_dropped
+
+
+def main():
+    argparse = ArgumentParser()
+    argparse.add_argument('--mongo_user', help='user to connect to mongodb', required=True)
+    argparse.add_argument('--mongo_password', help='password to connect to mongodb', required=True)
```
tcezard, line 101: I would make the user/password non required so you can run test the script. on Embassy.

tasks/eva_2124/correct_assembly_for_study.py
```diff
@@ -0,0 +1,112 @@
...
+    if contig in by_ucsc:
+        return by_ucsc[contig]['genbank'], False
+
+    if contig in by_refseq and by_refseq[contig]['is_genbank_refseq_identical']:
+        return by_refseq[contig]['genbank'], False
+
+    if contig in by_genbank:
+        return contig, True
+
+    raise Exception('could not find synonym for contig {}'.format(contig))
```
tcezard, line 37: What is the option if this occurs. 
We will have to change the script so it can accept an assembly report from the command line ?

tcezard (commented):
This looks really good and I'm impressed by how quickly you put it all together. 
Left a few comments none of which are critical.

tasks/eva_2124/correct_contig_for_study.py
```diff
@@ -0,0 +1,171 @@
...
+        except Exception:
+            unreplaceable_variants += contig['count']
+            unreplaceable_contigs.add(contig['_id'])
+
+    if len(unreplaceable_contigs) > 0:
+        raise Exception(
+            'Aborting replacement (no changes were done).'
+            ' With the provided assembly_report, the next {} contigs (present in {} variants) can not be replaced: {}'
+            .format(len(unreplaceable_contigs), unreplaceable_variants, unreplaceable_contigs))
+    elif replaceable_contigs == 0:
```
sundarvenkata-EBI, line 94: Shouldn't this be replaceable_contigs == 0 && unreplaceable_contigs == 0 && len(already_genbank_contigs) > 0 ?

jmmut: the current condition is the same as your suggestion but also including the case of `replaceable_contigs == 0 && already_genbank_contigs == 0`, which means that there are no variants in the DB (because we have also checked that `len(unreplaceable_contigs) == 0`), so arguably we should also do the exception, abort and message questioning the parameters.

sundarvenkata-EBI (approved):

PR #8, opened 2020-08-25
EVA-2121 aggregated table for database load

tcezard:
Add script that search for the reference sequence existing in EVA's storage and if can't be found download from NCBI.
Also allocate each species to one temp-mongo database


tasks/eva_2121/aggregate_table.py
```diff
@@ -0,0 +1,226 @@
...
+        if source == 'DBSNP - filesystem':
+            only_unmapped_tax_id.append(taxid)
+        elif 'no' not in stay_unchanged:
+            unchanged_taxid.append(taxid)
+        elif source == 'EVA':
+            eva_only.append(taxid)
+        elif source == 'DBSNP':
+            dbsnps_only.append(taxid)
+        else:
+            mixture.append(taxid)
```
sundarvenkata-EBI, line 145: I don't see any such category in the sheet...

tasks/eva_2121/aggregate_table.py
```diff
@@ -0,0 +1,226 @@
...
+
+    # Iterate a second time to assign a temp mongodb per taxonomy
+    with open(output_taxonmomy_tsv, 'w') as open_output:
+        headers = ['taxonomy_id', 'scientific_name', 'temp_mongo', 'to_copy', 'number_variants_to_process', 'number_variants']
```
sundarvenkata-EBI, line 149: Suggest ```headers = ['taxonomy_id', 'scientific_name', 'tempmongo_instance', 'should_be_copied', 'num_variants_in_taxid', 'total_num_variants']```

tasks/eva_2121/aggregate_table.py
```diff
@@ -0,0 +1,226 @@
...
+            if row['Decision to include'] == 'yes':
+                data_per_taxid_and_assembly[(row['Taxid'], row['Assembly'])].append(row)
+                data_per_taxid[row['Taxid']].append(row)
+    return data_per_taxid, data_per_taxid_and_assembly
+
+
+def assign_tempmongo_host_round_robin(number_variants, total_number_of_variant):
+    global assigned_number_variant
+    assigned_number_variant += number_variants
+    instance_number = int((assigned_number_variant // ((total_number_of_variant + 1) / 10)) + 1)
```
sundarvenkata-EBI, line 62: I guess this ensures some equitable distribution of the load to different instances based on the number of variants?

tasks/eva_2121/aggregate_table.py
```diff
@@ -0,0 +1,226 @@
...
+                taxid,
+                scientific_name,
+                assembly,
+                ','.join({row['Source'] for row in rows}),
+                fasta,
+                report,
+                temp_mongo,
+                to_process,
+                number_variants,
+                str(count_variants(rows, False))
```
sundarvenkata-EBI, line 202: Given the logic in count_variants, won't the result be the same as number_variants?

tasks/eva_2121/aggregate_table.py
tcezard: They are categories per taxonomy to detect which taxonomy needs to be processed and which does not. the `mixture` is where this taxonomy has data coming from at least two of dbsnp, eva and unmapped.

tasks/eva_2121/aggregate_table.py
tcezard: Changed excepted for `number_variants_to_process` which I prefer over `num_variants_in_taxid`

tasks/eva_2121/aggregate_table.py
tcezard: Yes this is the idea albeit not necessarily a clever one. If there is a more sensible way to split the data across, please let me know.

tasks/eva_2121/aggregate_table.py
tcezard: No, they will be the same number when all the variants for will be processed. But will be different when row with unmapped data.

tasks/eva_2121/aggregate_table.py
sundarvenkata-EBI: While this automatic assignment of species to instances is fine for the most part, I think the species that share assemblies should be in the same instance (to facilitate clustering), correct?

tasks/eva_2121/aggregate_table.py
sundarvenkata-EBI: Also, could we include the dbsnp name since we use that in the release folder name?

tcezard: That a good point to bring up. we should discuss the naming convention currently used.

tasks/eva_2121/aggregate_table.py
tcezard: There should not be any species that share assemblies in this release. The release code would not support it.

tasks/eva_2121/aggregate_table.py
```diff
@@ -3,41 +3,64 @@
...
+
+
+def parse_application_properties(app_prop):
+    properties_dict = { }
+    with open(app_prop) as open_file:
+        for line in open_file:
+            if not line.strip() or line.strip().startswith('#'):
+                continue
+            sp_line = line.strip().split('=')
+            properties_dict[sp_line[0]] = sp_line[1]
```
sundarvenkata-EBI, line 26:
```suggestion
            properties_dict[sp_line[0].strip()] = sp_line[1].strip()
```

tasks/eva_2121/aggregate_table.py
```diff
@@ -3,41 +3,64 @@
...
+    return properties_dict
+
+
+def find_and_parse_properties_and_parse(properties_path):
```
sundarvenkata-EBI, line 30:
```suggestion
def find_and_parse_properties(properties_path):
```

tasks/eva_2121/aggregate_table.py
```diff
@@ -0,0 +1,286 @@
...
+    property_files = glob.glob(os.path.join(properties_path, '*', 'release_phase1', '*', 'application.properties'))
```
sundarvenkata-EBI, line 31: I would also add '*application.properties' because IIRC some properties files will have an assembly accession prepended to the name :(

sundarvenkata-EBI (approved):
Looks good. Left a few minor comments.

sundarvenkata-EBI (approved):

tasks/eva_2121/aggregate_table.py
tcezard: Done

PR #9, opened 2020-08-29
EVA-2130 - script to fix contig name

tcezard:
Add option to only check the contig names can be changed and not change them
Check if contig is genbank first so it does not break if genbank id is used in any other field.
Prevent cursor timeout in the main update loop


jmmut (approved):
the progress and the genbank changes look ok, but I don't understand how you prevent the cursor timeout with the unused variable `no_cursor_timeout=True` and the try-finally.

tcezard:
The  [documentation](https://api.mongodb.com/python/current/api/pymongo/collection.html) for option `no_cursor_timeout` says that it will never timeout on the server (not sure how TBH) and the try/except/finally is simply to ensure the cursor is properly closed at the end

jmmut:
oh, right, I misread, I see it's passed to the find() method, I thought it was in the prototype of the enclosing function.

PR #10, opened 2020-09-01
EVA-2130  - remove duplicates

tcezard:
Script to remove duplicated entries. It come on top of PR https://github.com/EBIvariation/eva-tasks/pull/9

### DIFF
--- a/tasks/eva_2130/remove_duplicates_in_study.py
+++ b/tasks/eva_2130/remove_duplicates_in_study.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+from argparse import ArgumentParser
+import pymongo
+from urllib.parse import quote_plus
+
+
+def find_duplicates_and_remove_them(mongo_user, mongo_password, mongo_host, mongo_database, assembly_accession,
+                                    contig_list, study_list, dry_run):
+    duplicates_to_remove_commands = []
+    with get_mongo_connection_handle_url(
+            username=mongo_user,
+            password=mongo_password,
+            host=mongo_host
+    ) as accessioning_mongo_handle:
+        sve_collection = accessioning_mongo_handle[mongo_database]["submittedVariantEntity"]
+        cursor = sve_collection.find({
+            "seq": assembly_accession, "contig": {"$in": contig_list}, "study": {"$in": study_list}
+        })
+        for record in cursor:
+            variants_with_accessions = list(accessioning_mongo_handle[mongo_database]["submittedVariantEntity"].find({'accession': record['accession']}))
+            if len(variants_with_accessions) != 1:
+                print('Found %s duplicates for accession %s' % (len(variants_with_accessions), record['accession']) )
+            for variant in variants_with_accessions:
+                if variant['contig'] != record['contig']:
+                    duplicates_to_remove_commands.append(
+                        pymongo.DeleteOne({'contig': variant['contig'], 'accession': variant['accession']})
+                    )
+        if dry_run:
+            print("Will remove %s variants " % len(duplicates_to_remove_commands))
+        else:
+            sve_collection.bulk_write(requests=duplicates_to_remove_commands, ordered=False)
+            print("Remove %s variants " % len(duplicates_to_remove_commands))
+
+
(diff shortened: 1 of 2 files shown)


tasks/eva_2130/remove_duplicates_in_study.py
```diff
@@ -0,0 +1,63 @@
...
+    duplicates_to_remove_commands = []
+    with get_mongo_connection_handle_url(
+            username=mongo_user,
+            password=mongo_password,
+            host=mongo_host
+    ) as accessioning_mongo_handle:
+        sve_collection = accessioning_mongo_handle[mongo_database]["submittedVariantEntity"]
+        cursor = sve_collection.find({
+            "seq": assembly_accession, "contig": {"$in": contig_list}, "study": {"$in": study_list}
+        })
```
jmmut, line 18: maybe do the same trick to avoid cursor timeouts? if this is a long running query it will benefit from it.

tasks/eva_2130/remove_duplicates_in_study.py
```diff
@@ -0,0 +1,63 @@
...
+        for record in cursor:
+            variants_with_accessions = list(accessioning_mongo_handle[mongo_database]["submittedVariantEntity"].find({'accession': record['accession']}))
+            if len(variants_with_accessions) != 1:
+                print('Found %s duplicates for accession %s' % (len(variants_with_accessions), record['accession']) )
+            for variant in variants_with_accessions:
+                if variant['contig'] != record['contig']:
```
jmmut, line 24: I expected some more checks to identify duplicates. This will delete multimap accessions too. I guess it's fine if we know this script shouldn't be run on dbsnp collections. This script will become obsolete if we ever store multimap variants from our remapping.

I suggest considering checking that `variant['contig']` and `record['contig']` are synonyms according to the assembly report, and also checking that the identifying fields are the same (asm, start, type). If this script is really meant to be run only for EVA-2130, then my suggestion might be overkill, though.

tasks/eva_2130/remove_duplicates_in_study.py
```diff
@@ -0,0 +1,63 @@
...
+                    duplicates_to_remove_commands.append(
+                        pymongo.DeleteOne({'contig': variant['contig'], 'accession': variant['accession']})
+                    )
+        if dry_run:
+            print("Will remove %s variants " % len(duplicates_to_remove_commands))
+        else:
+            sve_collection.bulk_write(requests=duplicates_to_remove_commands, ordered=False)
```
jmmut, line 31: I guess this is fine as it's only meant for cleaning up the small (wrong) remnants of the correct_contig_for_study.py script, but if this script is meant for big deletions then I don't know how this will perform.

jmmut (changes requested):
mostly suggestions but the cursor timeout one is probably quite important.

tasks/eva_2130/remove_duplicates_in_study.py
tcezard: The code in EVA task is only meant for this task and this task alone. Even if I wanted to reuse the same code I would copy/paste it. I can add the test on identifying fields but adding extra complexity of loading the assembly reports seems unnecessary.

tasks/eva_2130/remove_duplicates_in_study.py
tcezard: For a 1000 variants I don't think it will timeout but better safe than sorry.

tasks/eva_2130/remove_duplicates_in_study.py
tcezard: I think for removing 993 variant it should be fine.

jmmut (approved):

tasks/eva_2130/remove_duplicates_in_study.py
sundarvenkata-EBI: Just for future information, deleting entries from MongoDB is brutally slow and any issue that prompts us to delete more than a million entries must be inspected more carefully for alternative remediations.

sundarvenkata-EBI (approved):

PR #14, opened 2020-09-28
EVA-2185 - Add option to filter by contig name

tcezard:


jmmut (approved):
there's another opened PR with a copy of this script, but for RSs, should we do this change there too?

tcezard:
> there's another opened PR with a copy of this script, but for RSs, should we do this change there too?

Don't think it is necessary, unless we have an actual need for it.

sundarvenkata-EBI (approved):

PR #15, opened 2020-10-08
EVA-2156 - Add script to deprecate variants

tcezard:


tasks/eva_2156/deprecate_variant_from_contig.py
```diff
@@ -0,0 +1,63 @@
...
+
+from ebi_eva_common_pyutils.config_utils import get_mongo_uri_for_eva_profile
+from ebi_eva_common_pyutils.logger import logging_config as log_cfg
+
+logger = log_cfg.get_logger(__name__)
+
+
+def inactive_object(variant):
+    return {
+        "eventType": "UPDATED",
```
andresfsilva, line 13: We are deprecating the variants so using the "DEPRECATED" [event type](https://github.com/EBIvariation/accession-commons/blob/master/accession-commons-core/src/main/java/uk/ac/ebi/ampt2d/commons/accession/core/models/EventType.java) is more accurate.

tasks/eva_2156/deprecate_variant_from_contig.py
```diff
@@ -0,0 +1,69 @@
...
+
+def inactive_object(variant):
+    inactive_variant = deepcopy(variant)
+    inactive_variant['hashedMessage'] = inactive_variant.pop('_id')
+    return {
+        "eventType": "DEPRECATED",
+        "accession": variant['accession'],
+        "reason": "Removed: The contig on which the variant was mapped is not accessioned by INSDC.",
+        "inactiveObjects": [inactive_variant],
+        "createdDate": datetime.now()  # "2018-07-30T13:21:59.333Z"
```
sundarvenkata-EBI, line 22: Just for the convenience of tracking, shall we have the same timestamp for the entire batch of contigs that are being deprecated?

tasks/eva_2156/deprecate_variant_from_contig.py
```diff
@@ -0,0 +1,69 @@
...
+
+logger = log_cfg.get_logger(__name__)
+
+
+def inactive_object(variant):
+    inactive_variant = deepcopy(variant)
+    inactive_variant['hashedMessage'] = inactive_variant.pop('_id')
+    return {
+        "eventType": "DEPRECATED",
+        "accession": variant['accession'],
```
sundarvenkata-EBI, line 19: My hope is that we can count on PyMongo to preserve the fidelity of this numeric field to NumberLongs (both here and in the inner inactiveObjects document) as in
image.... You can verify this after running the script...

sundarvenkata-EBI (approved):
Looks good! Left a minor comment...

tasks/eva_2156/deprecate_variant_from_contig.py
tcezard: I had a quick check and can confirm that it is the case.

tasks/eva_2156/deprecate_variant_from_contig.py
tcezard: Changed it.

sundarvenkata-EBI (approved):

PR #17, opened 2020-10-29
EVA-2194 - Count RS id per species

tcezard:
Three scripts to count the rs is specific file type in a release by species


sundarvenkata-EBI (approved):

andresfsilva (approved):
Looks good, I would suggest merging `count_rs_for_release.sh` and `count_rs_for_release_txt.sh` as there is only one line that is changing based on the file type, the rest of the logic is the same. An argument to specify if it is a VCF or txt would work and it would be more maintainable.

PR #21, opened 2021-02-02
EVA-2287 Fix contigs in study PRJEB42582 (Chr prefix removed)

andresfsilva:


tasks/eva_2287/correct_contig_with_chr_removed.py
```diff
@@ -0,0 +1,54 @@
...
+def get_SHA1(variant_rec):
+    """Calculate the SHA1 digest from the seq, study, contig, start, ref, and alt attributes of the variant"""
+    h = hashlib.sha1()
+    keys = ['seq', 'study', 'contig', 'start', 'ref', 'alt']
+    h.update('_'.join([str(variant_rec[key]) for key in keys]).encode())
+    return h.hexdigest().upper()
+
+
+def correct(mongo_user, mongo_password, mongo_host, mongo_database='eva_accession_sharded'):
+    with pymongo.MongoClient(mongo_host, username=mongo_user, password=mongo_password) as mongo_handle:
```
sundarvenkata-EBI, line 17: Please use method from pyutils to get handle without specifying the password in command line.

tasks/eva_2287/correct_contig_with_chr_removed.py
```diff
@@ -0,0 +1,54 @@
...
+        sve_collection = mongo_handle["eva_accession_sharded"]["submittedVariantEntity"]
+        filter_criteria = {'seq': 'GCA_002742125.1', 'study': 'PRJEB42582'}
+        cursor = sve_collection.find(filter_criteria)
+        insert_statements = []
+        drop_statements = []
+        number_of_variants_to_replace, number_of_variants_to_replace = 10, 10
```
sundarvenkata-EBI, line 23: These 2 seem to be the same...

tcezard (approved):

tasks/eva_2287/correct_contig_with_chr_removed.py
andresfsilva: Done. 
I was not using pyutils because I didn't want the dependency in this script that is intended to run just once but I agree passing the password in the command line is no a good practise. Tried using `get_mongo_uri_for_eva_profile` that would simplify things in this case but I wrote a test that uses a local mongo without authentication but that is not correctly handled by the functions in pyutils. Probably something we should fix in that library.

tasks/eva_2287/correct_contig_with_chr_removed.py
andresfsilva: Yes, it was in fact the same. Removed

tasks/eva_2287/correct_contig_with_chr_removed.py
apriltuesday: you could potentially patch the `get_mongo_uri_for_eva_profile` method in the tests to return a uri for your local mongo instance?  I don't think it matters much for this script but just a thought.

apriltuesday (approved):

sundarvenkata-EBI (approved):

tasks/eva_2287/correct_contig_with_chr_removed.py
andresfsilva: Thanks! Change done

PR #25, opened 2021-03-03
EVA-2323 - Script to dump basic json corresponding to the studies found in EVA

tcezard:
Simple script that translate the content of the EVAPRO database to JSON or get the same data from the all studies endpoint

### DIFF
--- a/tasks/eva_2323/dump_studies_json.py
+++ b/tasks/eva_2323/dump_studies_json.py
@@ -1,6 +1,6 @@
 import json
 from argparse import ArgumentParser
-from functools import cached_property
+from cached_property import cached_property
 
 import psycopg2
 from ebi_eva_common_pyutils.config_utils import get_properties_from_xml_file
--- a/tasks/eva_2323/dump_studies_json.py
+++ b/tasks/eva_2323/dump_studies_json.py
@@ -1,5 +1,7 @@
 import json
 from argparse import ArgumentParser
+
+import requests
 from cached_property import cached_property
 
 import psycopg2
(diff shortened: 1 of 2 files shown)


tasks/eva_2323/dump_studies_json.py
```diff
@@ -0,0 +1,106 @@
...
+            'entries': entries
+        }
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('--property_file')
+    args = parser.parse_args()
+    if args.property_file:
+        print(StudyDumper(args.property_file).dump_from_database())
```
sundarvenkata-EBI, line 100: I would suggest not using "development" as the default source because sometimes we might have data that has not yet been synchronized to PROD.

sundarvenkata-EBI (approved):

tasks/eva_2323/dump_studies_json.py
```diff
@@ -24,7 +24,7 @@ def connect(self):
         conn = psycopg2.connect(pg_url, user=pg_user, password=pg_pass)
         return conn
 
-    def dump_from_database(self):
+    def dump_from_database(self, properties_file, stage):
```
sundarvenkata-EBI, line 27: These two additional arguments don't seem to be used inside the function...

tcezard: That's true. I forgot to remove them.

andresfsilva (approved):
It looks good, you probably already know this but I would just mention that for now it is ok to have the option to use either the API or go directly to the DBs but at some point we will need to either choose only one or modify the `dump_from_database` to have the same logic as the web services so we end up with the same information independently of the source

PR #27, opened 2021-03-26
EVA-2357 Populate RS and SS IDs to the variant warehouse

andresfsilva:


tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+            variants_collection = mongo_handle[db_name]["variants_2_0"]
+            logging.info(f"Querying variants from variant warehouse, database {db_name}")
+            variants_cursor = get_variants_from_variant_warehouse(variants_collection)
+            hash_to_variant_id = {}
+            update_statements = []
+            try:
+                for variant_query_result in variants_cursor:
+                    # One variant in the variant warehouse can include more than one study, which means there can be
+                    # more than one submitted variant per variant warehouse variant
+                    get_ids(assembly, contig_synonym_dictionaries, variant_query_result, hash_to_variant_id)
```
sundarvenkata-EBI, line 173: To avoid implicit mutation of the passed dict, I suggest 
```suggestion
                   hash_to_variant_id = get_ids(assembly, contig_synonym_dictionaries, variant_query_result, hash_to_variant_id)
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    # Get SS ID from accessioning DB
+    sve_collection = mongo_handle[mongo_accession_db]['submittedVariantEntity']
+    sve_filter = {"_id": {"$in": list(sve_hashes)}}
+    sve_projection = {"_id": 1, "accession": 1, "rs": 1}
+    cursor_accessioning = sve_collection.find(sve_filter, projection=sve_projection)
+    hash_to_accession_info = {}
+    for sve in cursor_accessioning:
+        sve_hash = sve.get("_id")
+        ss_accession = sve.get('accession')
+        rs_accession = sve.get('rs')
```
sundarvenkata-EBI, line 47: I am hoping this won't throw an error when the rs attribute is null or not present...

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+            hash_to_variant_id = {}
+            update_statements = []
+            try:
+                for variant_query_result in variants_cursor:
+                    # One variant in the variant warehouse can include more than one study, which means there can be
+                    # more than one submitted variant per variant warehouse variant
+                    get_ids(assembly, contig_synonym_dictionaries, variant_query_result, hash_to_variant_id)
+                sve_hashes = hash_to_variant_id.keys()
+                logging.info(f"Querying accessioning database for assembly {assembly}")
+                hash_to_accession_info = get_from_accessioning_db(mongo_handle, mongo_accession_db, sve_hashes)
```
sundarvenkata-EBI, line 176: Similar to how you batch update requests below, I suggest batching lookups too so that we don't send out too many lookup requests at once...

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+                update_statements.extend(generate_update_statement(hash_to_variant_id, hash_to_accession_info))
+            except Exception as e:
+                print(traceback.format_exc())
+                raise e
+            finally:
+                variants_cursor.close()
+        logging.info(f"Updating database {db_name} variant warehouse with ss and rs ids")
+        result_update = variants_collection.bulk_write(requests=update_statements, ordered=False)
```
sundarvenkata-EBI, line 184: Could you refer to [this documentation](https://docs.mongodb.com/v4.0/reference/method/db.collection.bulkWrite/#bulk-write-with-write-concern) to bulk write with a majority write concern and a generous timeout so that we don't get cursor timeout errors?

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
+import hashlib
+import argparse
+import pymongo
+import traceback
+import logging
```
sundarvenkata-EBI, line 5: Suggest using logger from pyutils logging_cfg module since it comes with timestamps too...

tasks/eva_2357/populate_ids.py
sundarvenkata-EBI: Oh, wait... I am a bit confused. We are batching lookups but it looks like we are collecting the entire set of IDs in memory. I am not sure if this will pan out for large databases...

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+            logging.info(f"Querying variants from variant warehouse, database {db_name}")
+            variants_cursor = get_variants_from_variant_warehouse(variants_collection)
+            hash_to_variant_id = {}
+            update_statements = []
+            try:
+                for variant_query_result in variants_cursor:
+                    # One variant in the variant warehouse can include more than one study, which means there can be
+                    # more than one submitted variant per variant warehouse variant
+                    get_ids(assembly, contig_synonym_dictionaries, variant_query_result, hash_to_variant_id)
+                sve_hashes = hash_to_variant_id.keys()
```
sundarvenkata-EBI, line 174: Is **hash_to_variant_id** an in-memory dict of all the hashes determined from the variant warehouse? Will this scale for large databases?

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+import logging
+from ebi_eva_common_pyutils.config_utils import get_mongo_uri_for_eva_profile
+
+
+def generate_update_statement(hash_to_variant_ids, hash_to_accession_info):
+    variant_to_ids = {}
+    update_statements = []
+    for sve_hash, accessions in hash_to_accession_info.items():
+        ss_accession = f"ss{accessions.get('ss')}"
+        rs_accession = f"rs{accessions.get('rs')}" if accessions.get('rs') else accessions.get('rs')
```
sundarvenkata-EBI, line 14: Suggest hardcoding else to None...
```suggestion
        rs_accession = f"rs{accessions.get('rs')}" if accessions.get('rs') else None
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+        if variant_to_ids.get(variant_id):
+            variant_to_ids[variant_id].add(ss_accession)
+            if rs_accession:
+                variant_to_ids[variant_id].add(rs_accession)
+        else:
+            ids = set()
+            ids.add(ss_accession)
+            if rs_accession:
+                ids.add(rs_accession)
+            variant_to_ids[variant_id] = ids
```
sundarvenkata-EBI, line 27: Could this be shortened to?
```suggestion
            variant_to_ids[variant_id] = [ss_accession] + [rs_accession] if rs_accession else []
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    by_genbank = dict()
+    by_refseq = dict()
+    by_ucsc = dict()
+    with open(assembly_report_file, 'r') as f:
+        for line in f:
+            if line.startswith('#'):
+                continue
+
+            columns = line.strip().split('\t')
+            is_chromosome = columns[1] == 'assembled-molecule' and columns[3] == 'Chromosome'
```
sundarvenkata-EBI, line 79:
```suggestion
            is_chromosome = columns[1] == 'assembled-molecule' and columns[3].lower() == 'chromosome'
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    warehouse as value
+    """
+    for file in variant_query_result['files']:
+        study = file['sid']
+        start = variant_query_result['start']
+        ref = variant_query_result['ref']
+        alt = variant_query_result['alt']
+        id_variant_warehouse = variant_query_result['_id']
+        try:
+            genbank_chr, was_already_genbank = get_genbank(contig_synonym_dictionaries, variant_query_result['chr'])
```
sundarvenkata-EBI, line 135: We don't seem to use **was_already_genbank** anywhere?

sundarvenkata-EBI (commented):
Looks good! I have left a few comments.

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+        update_statements.append(pymongo.UpdateOne(query, update))
+    return update_statements
+
+
+def get_from_accessioning_db(mongo_handle, mongo_accession_db, sve_hashes):
+    # Get SS ID from accessioning DB
+    sve_collection = mongo_handle[mongo_accession_db]['submittedVariantEntity']
+    sve_filter = {"_id": {"$in": list(sve_hashes)}}
+    sve_projection = {"_id": 1, "accession": 1, "rs": 1}
+    cursor_accessioning = sve_collection.find(sve_filter, projection=sve_projection)
```
tcezard, line 42:
```suggestion
    cursor_accessioning = sve_collection.find(sve_filter, projection=sve_projection, no_cursor_timeout=True)
```

tasks/eva_2357/populate_ids.py
tcezard: It should not. This would have been converted to `dict`

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    hash_to_accession_info = {}
+    for sve in cursor_accessioning:
+        sve_hash = sve.get("_id")
+        ss_accession = sve.get('accession')
+        rs_accession = sve.get('rs')
+        hash_to_accession_info[sve_hash] = {"ss": ss_accession}
+        # Get RS ID is variant is clustered
+        if rs_accession:
+            hash_to_accession_info[sve_hash]['rs'] = rs_accession
```
tcezard, line 51: Why create the complex second `dict` layer where you could just add the `ss` and `rs` prefix here and keep them in a list ?

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    return hash_to_accession_info
+
+
+def get_variants_from_variant_warehouse(variants_collection):
+    projection = {"_id": 1, "files.sid": 1, "chr": 1, "start": 1, "ref": 1, "alt": 1, "type": 1}
+    return variants_collection.find({}, projection=projection)
```
tcezard, line 57:
```suggestion
    return variants_collection.find({}, projection=projection, no_cursor_timeout=True)
```

tcezard: This. query could potentially return a large number of variant. Is it worth batching them  with `.limit()` and `.skip()` 
Maybe a question for @sundarvenkata-EBI

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Get stats from variant warehouse', add_help=False)
+    parser.add_argument("--private-config-xml-file", help="ex: /path/to/eva-maven-settings.xml", required=True)
+    parser.add_argument("--dbs-to-populate-list",
+                        help="Full path to file with list of pairs (database, assembly) to populate the ids field "
+                             "with SS and RS ids (ex: /path/to/dbs/to/populate.txt)", required=True)
+    args = parser.parse_args()
+    populate_ids(args.private_config_xml_file, args.dbs_to_populate_list)
```
tcezard, line 196: I would suggest to implement `assert_all_contigs_can_be_replaced` a like in https://github.com/EBIvariation/eva-tasks/blob/ea411e706a939a70a8f54f4496bc410627f5a20b/tasks/eva_2124/correct_contig_for_study.py#L68

Last thing we want is to have the update half way through because there are missing chromosome name in the assembly report

tasks/eva_2357/populate_ids.py
tcezard: I would go one step further and not mutate the dict at all in the function
```suggestion
                    hash_to_variant_id.update(get_ids(assembly, contig_synonym_dictionaries, variant_query_result))
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+        return by_refseq[contig]['genbank'], False
+    raise Exception(f"Could not find synonym for contig {contig}")
+
+
+def get_SHA1(variant_id):
+    """Calculate the SHA1 digest from the seq, study, contig, start, ref, and alt attributes of the variant"""
+    return hashlib.sha1(variant_id.encode()).hexdigest().upper()
+
+
+def get_ids(assembly, contig_synonym_dictionaries, variant_query_result, hash_to_variant_id):
```
tcezard, line 123:
```suggestion
def get_ids(assembly, contig_synonym_dictionaries, variant_query_result):
```

tasks/eva_2357/populate_ids.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    Get a dictionary with the submitted variant hash as key and the variant id (chr_start_ref_alt) from the variant
+    warehouse as value
+    """
+    for file in variant_query_result['files']:
```
tcezard, line 128:
```suggestion
    hash_to_variant_id = {}
    for file in variant_query_result['files']:
```

tcezard (changes requested):

tasks/eva_2357/populate_ids.py
sundarvenkata-EBI: Yes, this should be batched. I have noted it in another comment.

tasks/eva_2357/populate_ids.py
andresfsilva: You are right, I removed it

tasks/eva_2357/populate_ids.py
andresfsilva: changed to use batches to process the query results

tasks/eva_2357/populate_ids.py
andresfsilva: But we need to keep the record of what variant those ss/rs ids belong to so the update statements can be constructed

tasks/eva_2357/populate_ids.py
andresfsilva: At the moment it would not crash. The current behaviour is to check if the contig has a genbank equivalent, if not if would skip it and log the message as a warning
```
[2021-Apr-07 15:30:55][tasks.eva_2357.populate_ids][WARNING] Could not find synonym for contig NC_000001.1 in variant NC_000001.1_2000_A_T
```
The other approach is checking if there is one contig that is not replaceable and abort updating the whole species if there is one. We can discuss that

tasks/eva_2357/populate_ids.py
tcezard: You need 2 things for the update statement:
 - The `sve_hash`
 - The list of ids you want to add
There is no need to separate the ss and rs ids in two different dicts since they will be added back to the same set.

tasks/eva_2357/populate_ids.py
tcezard: If we're not crashing when a chromosome cannot be resolved then we risk missing them out completely.
Even if we check the warnings, we will have to rerun the entire script with an updated assembly report which would be wasteful because most of the variants would have been updated already.

sundarvenkata-EBI (approved):

sundarvenkata-EBI (approved):

tcezard (approved):
For the contig name check I would have preferred a separate check that will use the aggregation pipeline to retrieve the chromosome names. They can then be checked against the assembly report before any changes are attempted. 
This is how [the equivalent script](https://github.com/EBIvariation/eva-tasks/blob/9c3b682e5fb6b0d57c9cf4d1fadf8eb24710aca0/tasks/eva_2124/correct_contig_for_study.py#L68) was working in EVA-2124. 
This can be done in a separate tasks though

PR #28, opened 2021-04-05
Analysis of covid19DP first batch

tcezard:


tasks/eva_2373/covid19dp_analysis.Rmd
```diff
@@ -0,0 +1,130 @@
+---
```
andresfsilva, line 1: Please change the file extension to `.md` to show it as a markdown

tasks/eva_2373/extract_AF.py
```diff
@@ -0,0 +1,27 @@
+from argparse import ArgumentParser
+
+import pysam
+
+
+def process_files(vcf_files_list):
+    """Extract Basic info from all the VCF files provided to populate a tsv file"""
```
andresfsilva, line 7: How are you populating the tsv? with the command line command using the messages being printed?

andresfsilva (approved):
This and the report look very good. I guess this is not intended to be run multiple times so we should not care about  hardcoding paths from your machine.

tasks/eva_2373/covid19dp_analysis.Rmd
```diff
@@ -0,0 +1,182 @@
...
+data_per_variant[data_per_variant$variantType=="INDEL" & data_per_variant$max_AF < 0.01 & data_per_variant$total>1000,]
+
+```
+
+
+## Distribution per sample
+
+```{r echo=FALSE, results="hide", message=FALSE}
+
+ggplot(data_per_sample, aes(total)) + geom_histogram(binwidth = 10) + xlab("Number of sample") + ylab("Number of variant") + facet_wrap( ~ variantType, scale="free_y")
```
apriltuesday, line 119: just a reminder about the axis labelling issue we discussed!

tasks/eva_2373/extract_AF.py
tcezard: It was through the `print` but I've changed it to an explicit output file

tasks/eva_2373/covid19dp_analysis.Rmd
tcezard: Done

apriltuesday (approved):

PR #30, opened 2021-04-19
EVA-2385 - Sharding report for Mongo DB's Collection

nitin-ebi:


nitin-ebi:
This PR is with reference to [EVA-2385](https://www.ebi.ac.uk/panda/jira/browse/EVA-2385)

tasks/eva_2385/sharding_report.py
```diff
@@ -0,0 +1,122 @@
...
+            collection = ns.split(".")[1]
+
+            logger.info(f"reading collection info for db({database}) and collection({collection})")
+            collection_info = mongo_client[database].command("collstats", collection)
+
+            collection_size = int(collection_info["size"])
+            result[ns] = {
+                "database": database,
+                "collection": collection,
+                "collection_size": collection_size,
```
sundarvenkata-EBI, line 69: Could you please convert the collection sizes to GB for human readability?

nitin-ebi: I had actually started with that, but for dbs with very small size, it was showing as 0, so I reverted it. 
If showing 0 for small db are fine, we can change this or if you want we can show size in both Bytes and GB by adding another column for GB in the existing structure.

sundarvenkata-EBI: ok, that sounds good!

tasks/eva_2385/sharding_report.py
```diff
@@ -0,0 +1,122 @@
...
+        database = value["database"]
+        collection = value["collection"]
+        collection_size = value["collection_size"]
+        for shard_key in sorted(value["shards"].keys()):
+            shard = value["shards"][shard_key]
+            data.append([database, collection, collection_size, shard_key, shard["shard_size"], shard["no_of_chunks"]])
+
+    logger.info("creating sharding report. writing data to csv file")
+    df = pandas.DataFrame(data, columns=header)
+    df.to_csv(path.join(report_dir, 'Shard_Report.csv'), index=False)
```
tcezard, line 100: Don't want to nitpick but adding a big dependency like panda to essentially write a csv file seems like a waste. 
Because it's a one off script I don't think it matters but we should avoid this in general.

tcezard (approved):

andresfsilva (approved):

tasks/eva_2385/sharding_report.py
nitin-ebi: added size in GB for collection

tasks/eva_2385/sharding_report.py
nitin-ebi: removed panda - code updated with csv module now.

PR #31, opened 2021-04-20
EVA-2389 - Script for multi-stage vertical concatenation of VCF files

sundarvenkata-EBI:
* Leverage non-linear pipeline model developed in [this PR](https://github.com/EBIvariation/eva-common-pyutils/pull/16) to script multi-stage vertical concatenation of VCF files.

* Suggest using https://github1s.com/EBIvariation/eva-tasks/pull/31/ to review the PR for proper rendering of ASCII art :)

* Usage noted [here](https://www.ebi.ac.uk/panda/jira/browse/EVA-2389?focusedCommentId=357490&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-357490)


tasks/eva_2389/run_vcf_vertical_concat_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+        process = NextFlowProcess(process_name=concat_stage_batch_name,
+                                  command_to_run=str(PythonProcess(vcf_vertical_concat,
+                                                                   {"files-to-concat-list": files_to_concat_list,
+                                                                    "concat-processing-dir": concat_processing_dir,
+                                                                    "output-vcf-file": output_vcf_file,
+                                                                    "bcftools-binary": bcftools_binary},
+                                                                   log_file=log_file_name
+                                                                   )
+                                                     )
+                                  )
```
apriltuesday, line 87: I don't quite get why we need to kick off a python process in the nextflow rather than just running bcftools concat/index directly from bash... maybe I'm missing something?

tasks/eva_2389/vcf_vertical_concat.py
```diff
@@ -0,0 +1,80 @@
...
+        self.bcftools_binary = bcftools_binary
+
+    def vertical_concat(self):
+        os.makedirs(os.path.dirname(self.output_vcf_file), exist_ok=True)
+        run_command_with_output(f"Running bcftools concat with the file list {self.files_to_concat_list}...",
+                                f"{self.bcftools_binary} concat "
+                                "--allow-overlaps --remove-duplicates "
+                                f"--file-list {self.files_to_concat_list} -o {self.output_vcf_file} -O z")
+        run_command_with_output(f"Running tabix on the output VCF {self.output_vcf_file}...",
+                                f"{self.bcftools_binary} index --csi {self.output_vcf_file}")
```
apriltuesday, line 53: Should this be `--tbi` or should the comment be modified?

tasks/eva_2389/run_vcf_vertical_concat_pipeline.py
sundarvenkata-EBI: At this point the concat process is just running "bcftools concat" but we will have to eventually enhance this process because the concat process [does not seem to properly concat the INFO fields](https://www.ebi.ac.uk/panda/jira/browse/EVA-2317?focusedCommentId=353543&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353543).

tasks/eva_2389/vcf_vertical_concat.py
sundarvenkata-EBI: No, I explicitly put in CSI because [CSI indices support longer genomes](http://www.htslib.org/doc/tabix.html)

tasks/eva_2389/run_vcf_vertical_concat_pipeline.py
apriltuesday: Using Nextflow makes perfect sense, I was asking about the first part - thanks for the explanation!

andresfsilva (approved):

tasks/eva_2389/vcf_vertical_concat.py
apriltuesday: Sure, I guess the comment could be changed then to match (not a big deal though).

apriltuesday (approved):

tasks/eva_2389/run_vcf_vertical_concat_pipeline.py
```diff
@@ -0,0 +1,167 @@
...
+import sys
+
+from ebi_eva_common_pyutils.logger import logging_config
+from ebi_eva_common_pyutils.nextflow import NextFlowPipeline, NextFlowProcess
+from tasks.eva_2389 import vcf_vertical_concat
+
+logger = logging_config.get_logger(__name__)
+
+
+class PythonProcess:
```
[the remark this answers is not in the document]
sundarvenkata-EBI: Fair point! Addressed.

tasks/eva_2389/vcf_vertical_concat.py
sundarvenkata-EBI: Addressed!

tcezard (approved):
This is well written and well tested, I have the same reservation as last time that this layer of abstraction is going to get incrementally more complex with higher maintenance cost.
I'm also worried that this create a second way of implementing Nextflow workflow which will result in different usage across the code base based on the programmer's preference. 
I'm happy for us to us this solution but I want us to do it as a conscious choice.

sundarvenkata-EBI:
> This is well written and well tested, I have the same reservation as last time that this layer of abstraction is going to get incrementally more complex with higher maintenance cost.
> I'm also worried that this create a second way of implementing Nextflow workflow which will result in different usage across the code base based on the programmer's preference.
> I'm happy for us to us this solution but I want us to do it as a conscious choice.

The Nextflow model provides a way to dynamically generate pipelines. I agree that it should not be reached for every time someone has to run a Nextflow pipeline. In this specific task, generating dynamic pipeline **_is a_** conscious decision since we generate the merge steps for each level at runtime. I have added the rationale for why the model was created in eva-common-pyutils if you would like to take a look.

PR #33, opened 2021-04-27
EVA-2399 Script to validate counts between DBs in Mongo 3.2 and 4.0 cluster

nitin-ebi:


tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,136 @@
...
+            no_of_documents_in_src = get_documents_count_for_collection(mongo_source, db, coll)
+            no_of_documents_in_dest = get_documents_count_for_collection(mongo_dest, db, coll)
+
+            if (no_of_documents_in_src == 0) and (coll not in source_collections):
+                logger.info(f"collection {coll} does not exist in database {db} in mongo source")
+                no_of_documents_in_src = -1
+
+            if (no_of_documents_in_dest == 0) and (coll not in dest_collections):
+                logger.info(f"collection {coll} does not exist in database {db} in dest source")
+                no_of_documents_in_dest = -1
```
tcezard, line 45: Why not do these test of existence of the collections  before the call for `get_documents_count_for_collection` ?
Otherwise you will access (and retry 3 times) the database for nothing.

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,136 @@
...
+
+            if no_of_documents_in_src != no_of_documents_in_dest and \
+                    no_of_documents_in_src != -1 and no_of_documents_in_dest != -1:
+                logger.info(f"no of documents in collection {coll} of database {db} does not match")
```
tcezard, line 49: Could you add the actual number to the log statement?

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: If collection does not exist, it does not fail and retry but simply returns 0. 
But agreed, it can be avoided altogether. Have updated the code to check and skip if collection does not exist in any instance.

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: added

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,138 @@
...
+    count_validation_res_list = []
+
+    for db in database_list:
+        mongo_source.db_name = db
+        mongo_dest.db_name = db
+        source_collections = mongo_source.get_collection_names()
+        dest_collections = mongo_dest.get_collection_names()
+
+        if not source_collections and not dest_collections:
+            logger.info(f"database {db} does not exit in either of mongo instances")
```
sundarvenkata-EBI, line 28:
```suggestion
            logger.info(f"database {db} does not exist in either of mongo instances")
```

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,138 @@
...
+            continue
+
+        all_collections = sorted(set(source_collections).union(set(dest_collections)))
+
+        for coll in all_collections:
+            logger.info(f"fetching count for collection ({coll}) in database ({db})")
+
+            if coll not in source_collections:
+                logger.info(f"collection ({coll}) does not exist in database ({db}) in mongo source")
```
sundarvenkata-EBI, line 37: This should be flagged as an error!
```suggestion
                logger.error(f"collection ({coll}) does not exist in database ({db}) in mongo source")
```

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,138 @@
...
+                no_of_documents_in_src = -1
+            else:
+                no_of_documents_in_src = get_documents_count_for_collection(mongo_source, db, coll)
+
+            if coll not in dest_collections:
+                logger.info(f"collection ({coll}) does not exist in database ({db}) in mongo destination")
```
sundarvenkata-EBI, line 43: This should be flagged as an error!

```suggestion
                logger.error(f"collection ({coll}) does not exist in database ({db}) in mongo destination")
```

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,138 @@
...
+            else:
+                no_of_documents_in_dest = get_documents_count_for_collection(mongo_dest, db, coll)
+
+            if no_of_documents_in_src != no_of_documents_in_dest and \
+                    no_of_documents_in_src != -1 and no_of_documents_in_dest != -1:
+                logger.info(f"no of documents in collection ({coll}) of database ({db}) does not match : "
+                            f"source has ({no_of_documents_in_src}) documents whereas destination has ({no_of_documents_in_dest})")
+
+            count_validation_res_list.append(
+                (db, coll, no_of_documents_in_src, no_of_documents_in_dest, datetime.now()))
```
sundarvenkata-EBI, line 54: Since we intend to run this script for a set of databases, I would suggest using a single timestamp for a given run to avoid confusion.

tasks/eva_2399/mongo_count_validation.py
```diff
@@ -0,0 +1,138 @@
...
+
+    return count_validation_res_list
+
+
+@retry(logger=logger, tries=3, delay=2)
```
sundarvenkata-EBI, line 59: Could you also try backoff + jitter parameters? If the request fails because the database is busy with other things, hammering it back-to-back might be a bit dangerous.

sundarvenkata-EBI (changes requested):
Looks good! But left a few comments about issues that should be addressed.

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: updated

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: updated

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: updated

tasks/eva_2399/mongo_count_validation.py
nitin-ebi: updated

sundarvenkata-EBI (approved):

PR #36, opened 2021-05-06
EVA-2400 - Mongo DB Document Count Validation - Script Update

nitin-ebi:
**Current Behaviour :**
Currently the script tries to connect to two different mongo instances in order to validate the no of documents in corresponding db/collections.
**Changes in PR:**
This PR is to update the script so that it connects with one mongo instance at a time and log the no of documents in each db/collection only for that instance.
**Advantages:**

1. Can run separate instances of the script for separate mongo instances, simultaneously.
2. If there is an issue with one of the mongo instance (down, busy or throws errors), it will not hamper the counting for the other
3. Flexibility to run for only one mongo instance.
4. Can run on a suitable time for an instance (when the load is low). Doesn't have to consider any other instance's availability and load

Results are reported in the table as follows:
image


nitin-ebi:
**Current Behaviour :**
Currently the script tries to connect to two different mongo instances in order to validate the no of documents in corresponding db/collections.  
**Changes in PR:**
This PR is to update the script so that it connects with one mongo instance at a time and log the no of documents in each db/collection only for that instance. 
**Advantages:**
1. Can run separate instances of the script for separate mongo instances, simultaneously.
2. If there is an issue with one of the mongo instance (down, busy or throws errors), it will not hamper the counting for the other
3. Flexibility to run for only one mongo instance.
4. Can run on a suitable time for an instance (when the load is low). Doesn't have to consider any other instance's availability and load

sundarvenkata-EBI (approved):

tcezard (approved):
This looks good however, it needs to be rebased and the conflicts needs to be resolved

PR #39, opened 2021-06-11
EVA-2464 Correct contig error in study PRJEB43246

nitin-ebi:


tasks/eva_2461/correct_contig_error_in_study.py
```diff
@@ -0,0 +1,65 @@
...
+    keys = ['seq', 'study', 'contig', 'start', 'ref', 'alt']
+    h.update('_'.join([str(variant_rec[key]) for key in keys]).encode())
+    return h.hexdigest().upper()
+
+
+def replace_with_correct_contig(mongo_source):
+    correct_contig = 'AF034253.1'
+    sve_collection = mongo_source.mongo_handle[mongo_source.db_name]["submittedVariantEntity"]
+    filter_criteria = {'seq': 'GCA_000003025.4', 'study': 'PRJEB43246', 'contig': 'M'}
+    cursor = sve_collection.find(filter_criteria)
```
sundarvenkata-EBI, line 22: Even though we know the number of documents is small,  Mongo is trying to find a needle in a haystack, I would suggest providing a parameter for cursor timeout with a generous value here.

nitin-ebi:
Code adapted from existing script [correct_contig_with_chr_removed](https://github.com/EBIvariation/eva-tasks/blob/master/tasks/eva_2287/correct_contig_with_chr_removed.py)

tasks/eva_2461/correct_contig_error_in_study.py
```diff
@@ -0,0 +1,65 @@
...
+    try:
+        for variant in cursor:
+            original_id = get_SHA1(variant)
+            assert variant['_id'] == original_id, "Original id is different from the one calculated %s != %s" % (
+                variant['_id'], original_id)
+            variant['contig'] = correct_contig
+            variant['_id'] = get_SHA1(variant)
+            insert_statements.append(pymongo.InsertOne(variant))
+            drop_statements.append(pymongo.DeleteOne({'_id': original_id}))
+        result_insert = sve_collection.bulk_write(requests=insert_statements, ordered=False)
```
sundarvenkata-EBI, line 36: Suggest using majority write concern here...

tasks/eva_2461/correct_contig_error_in_study.py
```diff
@@ -0,0 +1,65 @@
...
+        total_inserted += result_insert.inserted_count
+        result_drop = sve_collection.bulk_write(requests=drop_statements, ordered=False)
```
sundarvenkata-EBI, line 38: Suggest using majority write concern here...

sundarvenkata-EBI (changes requested):
Looks good! Left a few comments regarding read/write concerns and timeouts.

tasks/eva_2461/correct_contig_error_in_study.py
tcezard: I would advise to pass `no_cursor_timeout=True` see [doc](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.find)

tcezard (approved):

tasks/eva_2461/correct_contig_error_in_study.py
nitin-ebi: updated

tasks/eva_2461/correct_contig_error_in_study.py
nitin-ebi: updated

tasks/eva_2461/correct_contig_error_in_study.py
nitin-ebi: updated

sundarvenkata-EBI (approved):

tasks/eva_2461/correct_contig_error_in_study.py
```diff
@@ -0,0 +1,68 @@
...
+    keys = ['seq', 'study', 'contig', 'start', 'ref', 'alt']
+    h.update('_'.join([str(variant_rec[key]) for key in keys]).encode())
+    return h.hexdigest().upper()
+
+
+def replace_with_correct_contig(mongo_source):
+    correct_contig = 'AF034253.1'
+    sve_collection = mongo_source.mongo_handle[mongo_source.db_name]["submittedVariantEntity"]
+    filter_criteria = {'seq': 'GCA_000003025.4', 'study': 'PRJEB43246', 'contig': 'M'}
+    cursor = sve_collection.find(filter_criteria, no_cursor_timeout=True)
```
sundarvenkata-EBI, line 23: Could you also check to see if you can use "majority" readConcern here?

nitin-ebi: updated

sundarvenkata-EBI (approved):

PR #40, opened 2021-06-18
EVA-2403 - Check contigs can be translated to Genbank

tcezard:


tasks/eva_2357/test/test_populate_ids.py
```diff
@@ -147,22 +149,26 @@ def test_populate_ids_fail(self, mock_get_mongo_uri_for_eva_profile):
...
-        self.assertCountEqual(variant['ids'], ['ss1'])
+        check_all_contigs(settings, databases_file, profile='localhost')
+
+    @patch('tasks.eva_2357.populate_ids.get_mongo_uri_for_eva_profile')
+    def test_check_all_contigs_raise(self, mock_get_mongo_uri_for_eva_profile):
+        logging.getLogger().setLevel(logging.DEBUG)
+        mock_get_mongo_uri_for_eva_profile.return_value = 'mongodb://127.0.0.1:27017'
+        settings = self.get_test_resource("settings.xml")
+        databases_file = self.get_test_resource("databases_errors.txt")
+        with pytest.raises(ValueError):
```
apriltuesday, line 169: If you do this you shouldn't need to import `pytest`...
```suggestion
        with self.assertRaises(ValueError):
```

tasks/eva_2406/tests/test_gather_release_species.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
apriltuesday, line 1: I think this and the next file must've slipped in from your other ticket!

apriltuesday (approved):

tasks/eva_2357/test/test_populate_ids.py
tcezard: Done

tasks/eva_2406/tests/test_gather_release_species.py
tcezard: Indeed 😊

andresfsilva (approved):

PR #44, opened 2021-07-01
EVA-2155 Script to fix MT contigs in PRJEB33693

andresfsilva:


tasks/eva_2155/correct_contig_error_in_study.py
```diff
@@ -0,0 +1,89 @@
...
+        print(traceback.format_exc())
+        raise e
+    finally:
+        cursor.close()
+    return total_inserted
+
+
+def delete_variants_wront_contig(mongo_source):
+    sve_collection = mongo_source.mongo_handle[mongo_source.db_name]["submittedVariantEntity"]
+    filter_criteria = {'seq': 'GCA_000298735.1', 'study': 'PRJEB33693', 'contig': 'OARMT'}
```
sundarvenkata-EBI, line 52: Since it has been a [year since the accessions have been published](https://www.ebi.ac.uk/panda/jira/browse/ELOAD-531?focusedCommentId=323989&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-323989), could you or @BaronKoy check with the submitter + FTP logs to understand if the accessions have been referenced anywhere (for example, in publications) so that we can determine the correct course of action ([merging](https://github.com/EBIvariation/eva-tasks/blob/5411bd9a1187140480efd04fa159bd0b4aad0123/tasks/eva_2146/merge_prjeb25004_redundant_ssids.groovy) might be preferable if they have already been used) for redundant variants.

sundarvenkata-EBI: Discussed offline and @andresfsilva confirmed that the OARMT accessions in PROD are not being used in the accessioning file. So the deletion can go ahead as planned.

sundarvenkata-EBI (approved):

nitin-ebi (approved):

PR #48, opened 2021-10-15
Restore deleted study by pulling metadata from EVAPRO back to EVADEV

tcezard:


sundarvenkata-EBI (approved):
Looks good! But could you please add context around the events/issues for specific studies that prompted us to write this script?

tcezard:
This was written to remediate a study deletion from EVADEV. The study still existed in EVAPRO and could be brought back  this way.

PR #50, opened 2021-11-09
EVA-2641: second part of remediation

apriltuesday:
Read annotations from the temporary collection created [here](https://github.com/EBIvariation/eva-tasks/pull/49) and write them to the appropriate place, using the same writers as in `eva-pipeline`.

Also includes some fixes to the clean-up script.

### DIFF
--- a/tasks/eva_2641/write_annotations.groovy
+++ b/tasks/eva_2641/write_annotations.groovy
@@ -1,0 +1,90 @@
+@Grab(group='uk.ac.ebi.eva', module='eva-pipeline', version='2.8-SNAPSHOT')
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.batch.item.ItemWriter
+import org.springframework.batch.item.support.CompositeItemWriter
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.data.mongodb.core.MongoOperations
+
+import uk.ac.ebi.eva.commons.models.metadata.AnnotationMetadata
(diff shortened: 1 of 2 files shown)


tasks/eva_2641/write_annotations.groovy
```diff
@@ -0,0 +1,102 @@
...
+        logger.info("VEP version {}, cache version {}", vepVersion, vepCacheVersion)
+
+        annotationItemWriter = annotationItemWriter(mongoOperations)
+        variantAnnotationItemWriter = variantAnnotationItemWriter(mongoOperations)
+        annotationWriter = compositeAnnotationItemWriter()
+
+        List<Annotation> annotations = readAnnotationsFromTempCollection()
+        logger.info("Read {} annotations", annotations.size())
+
+        annotationWriter.write(Collections.singletonList(annotations))
```
sundarvenkata-EBI, line 60: singletonList can hold only one element. Did you mean to put in unModifiableList here?

apriltuesday: It should be a one-element list containing a list of annotations...  In the pipeline I think the inner list would be a chunk of 200 annotations that are themselves written in batches, I wasn't sure if that was necessary in this script, as we're not actually passing variants through VEP or anything.

sundarvenkata-EBI: Got it... I see your point but let's keep the method signature as it is even if it means using a singletonList here...

sundarvenkata-EBI (approved):

tcezard (approved):

PR #56, opened 2022-01-06
EVA-2706 - Discordant clustered variants 

tcezard:
Script to detect clustered variant discordant with their submitted variants

### DIFF
--- a/tasks/eva_2706/detect_discordant_cluster_variant.py
+++ b/tasks/eva_2706/detect_discordant_cluster_variant.py
@@ -18,6 +18,8 @@ def grouper(n, iterable, fillvalue=None):
 def compare(clustered_variants, submitted_variant_position_per_rs):
     error = 0
     for clustered_variant in clustered_variants:
+        if not clustered_variant:
+            continue
         pos = f'{clustered_variant.get("contig")}:{clustered_variant.get("start")}'
         if clustered_variant['accession'] not in submitted_variant_position_per_rs:
             logger.error(f'No submitted variant found for rs{clustered_variant["accession"]}')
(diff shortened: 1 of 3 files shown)


nitin-ebi (approved):

tasks/eva_2706/detect_discordant_cluster_variant.py
```diff
@@ -0,0 +1,89 @@
...
+
+def compare(clustered_variants, submitted_variant_position_per_rs):
+    error = 0
+    for clustered_variant in clustered_variants:
+        if not clustered_variant:
+            continue
+        pos = f'{clustered_variant.get("contig")}:{clustered_variant.get("start")}'
+        if clustered_variant['accession'] not in submitted_variant_position_per_rs:
+            logger.error(f'No submitted variant found for rs{clustered_variant["accession"]}')
+            error += 1
```
apriltuesday, line 26: Do we want to count these errors in this script, or just warn?  Feels like a separate issue from the position mismatch...

tasks/eva_2706/detect_discordant_cluster_variant.py
```diff
@@ -0,0 +1,89 @@
...
+        for sve in sve_cursor:
+            submitted_variant_position_per_rs[sve.get('rs')].add(f"{sve.get('contig')}:{sve.get('start')}")
+
+        nb_error += compare(clustered_variants, submitted_variant_position_per_rs)
+        logger.info(f'Processed {nb_clustered_variants}: Found {nb_error} errors')
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Delete declustered variants in dbsnpSubmittedVariantEntity Collection')
```
apriltuesday, line 74: should be updated...

apriltuesday (approved):

tasks/eva_2706/detect_discordant_cluster_variant.py
tcezard: I did count them as they seemed important and they would have the same impact on a merge/split as a misplaced ssid. This said I agree that they are a separate issue and we didn't spend much time on figuring out why they occur. 
I'll add another ticket for them in JIRA.

PR #63, opened 2022-03-02
Ipython notebook analysis of the amount of data downloaded from the FTP.

tcezard:


apriltuesday (commented):
This is cool, I'm excited about having some reproducible charts!  I didn't review the code much but some comments on the charts:

For the per-species release 1 donut chart, I wonder if it's worth using fewer than 20 categories since it's impossible to read the last few anyway?  Otherwise you might have to do some fiddling to get the labels for the small categories to be visible, or maybe do something more like [this second example](https://matplotlib.org/stable/gallery/pie_and_polar_charts/pie_and_donut_labels.html).

Also for both per-species donut plots, it looks like there's a species missing a label (pink wedge in the bottom/bottom left).

PR #64, opened 2022-03-14
EVA-2722: prepare file for already processed analysis of covid19 project

nitin-ebi:


tasks/eva_2722/prepare_covid19_processed_analysis_file.py
```diff
@@ -0,0 +1,101 @@
...
+        file_path = os.path.join(processed_file_directory, file)
+        if os.path.islink(file_path):
+            org_path = os.readlink(file_path)
+            logger.info(f"{file_path} is a symlink, reading files from original path {org_path}")
+            processed_analysis.update(get_processed_analysis_files(org_path))
+        elif os.path.isdir(file_path):
+            logger.info(f"{file_path} is a directory, reading files from sub-directories")
+            processed_analysis.update(get_processed_analysis_files(file_path))
+        else:
+            if "_filtered_vcf.gz" in file:
```
sundarvenkata-EBI, line 66: I am fine with doing this but have you validated that all of the filtered VCF file names adhere to this pattern?

tasks/eva_2722/prepare_covid19_processed_analysis_file.py
```diff
@@ -0,0 +1,101 @@
...
+    return response.json()
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Download analysis for processing from the Covid-19 DP project',
+                                     formatter_class=argparse.RawTextHelpFormatter, add_help=False)
+    parser.add_argument("--project", default='PRJEB45554', required=False,
+                        help="project from which analysis needs to be downloaded")
+    parser.add_argument("--batch-size", default=10000, required=False, help="batch size of ENA analysis download")
+    parser.add_argument("--processed-file-directory", default="/nfs/production3/eva/data/PRJEB45554/30_eva_valid/",
```
sundarvenkata-EBI, line 89: Since this is a public repo, suggest not using actual cluster file paths...

sundarvenkata-EBI (approved):

tasks/eva_2722/prepare_covid19_processed_analysis_file.py
nitin-ebi: removed default and made the parameter mandatory

tasks/eva_2722/prepare_covid19_processed_analysis_file.py
nitin-ebi: yes

tcezard (commented):
This looks ok but it makes the assumption that all the already processed files have an analysis in ENA. 
I'm not sure this is correct. 
Could you check that it is within the script  and count the number of file for which you did not find an analysis

nitin-ebi:
> This looks ok but it makes the assumption that all the already processed files have an analysis in ENA. I'm not sure this is correct. Could you check that it is within the script and count the number of file for which you did not find an analysis

added the check and count for files with no analysis in ENA

tcezard (approved):

PR #69, opened 2022-04-26
Avirup

itsroops:


tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,270 @@
...
+# Initializing the list to store all the keys and values for all the files
+all_keys=[]
+all_values=[]
+
+# Initializing the list to store the values of all taxonomy ids and assembly accessions
+all_tax_id=[]
+all_assembly=[]
+
+
+def generate_output(remapping_root_path, output_path):
```
tcezard, line 35: From the function name, we would assume that it will only create the output but it does create the entire data structure as well.
I think it makes sense to have a function to create the output but then you should extract the functionalities that retrieve the data to another function.

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,270 @@
...
+    keys_values = gather_counts_per_file(filename_eva)
+    keys_values_dbsnp = gather_counts_per_file(filename_dbsnp)
+
+    # Adding the values of dbsnp and eva with the common keys
+    for key in keys_values_dbsnp:
+        if key in keys_values:
+            keys_values_dbsnp[key] = keys_values_dbsnp[key] + keys_values[key]
+
+    # Concatenating the two dictionaries
+    keys_values_dbsnp = {**keys_values, **keys_values_dbsnp}
```
tcezard, line 168: This works but I find it confusing and the name are not very descriptive.
I would suggest:
```suggestion
    all_headers = list(set(keys_values) + set(keys_values_dbsnp))
    all_values = {}
    for key in set(keys_values) + set(keys_values_dbsnp):
        if key in keys_values:
            all_values[key] = keys_values_dbsnp.get(key, 0) + keys_values.get(key, 0)
```

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # Adding the taxonomy and assembly accession for a particular file
+    all_tax_id.append(taxid)
+    all_assembly.append(assembly_accession)
+
+    # Updating the final output columns
+    global all_columns
+    all_columns.extend(keys)
+    all_columns = list(dict.fromkeys(all_columns)
+    all_columns.sort()
```
tcezard, line 190: All these don't really belong in this function. 
It should be extracted out to the top level function.

tcezard (changes requested):
This code is used as a one off so it does not need to be perfect but it is better to get use to organise the code in independent function bits.

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: The code is reorganized and the suggestions are incorporated

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: I have created a separate function for generation of output

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: I have now used more descriptive names

tcezard: I see.
My proposal was more about the way you create the merge in two loop where I do it in in one using default values when the key is not present.

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
tcezard: You have simplified it indeed but it still contains both the data retrieval and the generation of the output in the same function.

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # Calling functions for gathering stats per file from eva and dbsnp data and storing the data in a dictionary
+    # as a key-value pair
+    columns_values = gather_counts_per_file(filename_eva)
+    columns_values_dbsnp = gather_counts_per_file(filename_dbsnp)
+
+    # Adding the values of dbsnp and eva with the common columns i.e. fields
+    for column in columns_values_dbsnp:
+        if column in columns_values:
+            columns_values_dbsnp[column] = columns_values_dbsnp[column] + columns_values[column]
```
apriltuesday, line 206: I find it a bit counter-intuitive to accumulate the values in `column_values_dbsnp` rather than the more generically named `column_values` (or into a third dict altogether)...

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,311 @@
...
+           spreadsheet
+
+    Output: It calls the extract_taxid_assembly() function to gather the taxonomy ids and the assembly accession
+            which in turn calls the gather_counts_per_tax_per_assembly() function for a particular taxonomy id
+            and a particular assembly to collect the statistics and generate the output spreadsheet
+    """
+
+    # Collecting the values of all_taxonomy ids and their corresponding assemblies
+    all_columns, all_columns_for_specific_taxid_assembly, all_values_for_specific_taxid_assembly, \
+    all_tax_id, all_assembly = extract_taxid_assembly(remapping_root_path)
```
apriltuesday, line 39:
```suggestion
        all_tax_id, all_assembly = extract_taxid_assembly(remapping_root_path)
```
(for readability)

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,311 @@
...
+        'valign': 'vcenter',
+        'fg_color': '#D7E4BC',
+        'text_wrap': 1})
+
+    # Considering maximum reasons of failure to be restricted to 26*3
+    letters = list(ascii_uppercase[:3])
+    all_letters = list(ascii_uppercase)
+    a_z = list(ascii_uppercase)
+
+    # Storing the Excel cell header subscripts in a list
```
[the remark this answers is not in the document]
itsroops: Thanks for the suggestions...I have incorporated them now

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,311 @@
...
+    # Returning the fields in a yaml file along with its value for a particular tax id and accession
+    return columns, values, taxid, assembly_accession
+
+
+def gather_counts_per_file(filename):
+    """
+    This function is used to load the yml files for both eva and dbsnp and gather the data from the key-value pairs
+    in the yml files
+
+    Input: The absolute filepaths and filenames of the eva and the dbsnp files
```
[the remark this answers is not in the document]
itsroops: Yes, it is a single filename

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: I have now made the relevant changes for  your review

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: I have now incorporated the changes in the code for review

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def extract_taxid_assembly(remapping_root_path):
+    """
+    This function is used to extract the full list of taxonomy ids and assembly accessions and pass it to
+    gather_counts_per_tax_per_assembly() function for a particular taxonomy id and assembly for gathering statistics and subsequently consolidates statistics for a holistic analysis
+
+    Input: It accepts the remapping full path for subsequent analysis
+
+    Output: It calls gather_counts_per_tax_per_assembly() function for a particular taxonomy id and
```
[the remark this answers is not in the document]
itsroops: Implementation details have been omitted in the revised code

tasks/eva_2835/gather_per_species_per_assembly_variant_remapping_counts.py
itsroops: I have made this change in the revised code

apriltuesday (approved):

tcezard (approved):

PR #72, opened 2022-05-25
EVA-2861 - Remediation in PROD

sundarvenkata-EBI:
* [Replace var with def to avoid weird compiler bug GROOVY-5961](https://github.com/EBIvariation/eva-tasks/pull/72/commits/e1e30c8406a7d6a20e49dd9040e0317f83c24845)
* [Process outstanding split candidates in remediation script](https://github.com/EBIvariation/eva-tasks/pull/72/commits/b661cbd38d3a0c384c4beb7873b1e084a444044b)

Usage [here](https://www.ebi.ac.uk/panda/jira/browse/EVA-2861?focusedCommentId=394900&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-394900)

### DIFF
--- a/tasks/eva_2861/remediate_duplicate_dbsnp_ss/src/main/groovy/uk/ac/ebi/eva/remediate_duplicate_dbsnp_ss/RemediateDuplicateDbsnpSs.groovy
+++ b/tasks/eva_2861/remediate_duplicate_dbsnp_ss/src/main/groovy/uk/ac/ebi/eva/remediate_duplicate_dbsnp_ss/RemediateDuplicateDbsnpSs.groovy
@@ -1,4 +1,4 @@
-package uk.ac.ebi.eva.remediate_duplicate_dbsnp_ss
+package uk.ac.ebi.eva.accession.clustering.batch.io
 
 @Grab(group = 'uk.ac.ebi.eva', module = 'eva-accession-core', version = '0.6.10-SNAPSHOT')
 @Grab(group = 'uk.ac.ebi.eva', module = 'eva-accession-clustering', version = '0.6.10-SNAPSHOT')
@@ -10,6 +10,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.batch.item.ItemWriter
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.CommandLineRunner
 
 import org.springframework.boot.SpringApplication
@@ -22,6 +23,7 @@ import org.springframework.data.mongodb.core.query.Query
 import org.springframework.retry.annotation.Backoff
 import org.springframework.retry.annotation.Retryable
 import org.springframework.stereotype.Component
+import uk.ac.ebi.eva.accession.clustering.batch.io.SSSplitWriter
 import uk.ac.ebi.eva.accession.clustering.configuration.BeanNames
 import uk.ac.ebi.eva.accession.clustering.configuration.InputParametersConfiguration
 import uk.ac.ebi.eva.accession.clustering.configuration.batch.io.SSSplitWriterConfiguration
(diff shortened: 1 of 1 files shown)


tasks/eva_2861/remediate_duplicate_dbsnp_ss/src/main/groovy/uk/ac/ebi/eva/remediate_duplicate_dbsnp_ss/RemediateDuplicateDbsnpSs.groovy
```diff
@@ -79,13 +113,33 @@ class RemediateDuplicateDbsnpSs implements CommandLineRunner {
             queryToGetNextBatchOfSS.addCriteria(where("_id").gt(lastSeenID))
         }
         queryToGetNextBatchOfSS = queryToGetNextBatchOfSS.with(Sort.by(Sort.Direction.ASC, "_id")).limit(1000)
-        List<? extends SubmittedVariantEntity> result = this.mongoTemplate.find(queryToGetNextBatchOfSS,
-                DbsnpSubmittedVariantEntity.class)
+        // For the sake of efficiency, get documents from DEV instance rather than the PROD instance
+        // because we have only imported a restricted subset of documents that share duplicate IDs in DEV
+        // see LoadDuplicateDbsnpSSToTestDB.groovy
```
tcezard, line 118: If I understand correctly the process will be:
- Get the files defined in EVA-2840, load the ssids/assemblies from it and copy the corresponding SubmittedVariants from prod to dev
- Read the Submitted Variants from dev and split them in prod.
I I understand correctly, this mean we will have to be careful to clear the dev database of the test remediation that already occurred.

sundarvenkata-EBI: #2 is not needed. As long as there is one of the duplicate IDs in DEV it will be fine since the SplitWriter ([autowired to PROD](https://github.com/sundarvenkata-EBI/eva-tasks/blob/b661cbd38d3a0c384c4beb7873b1e084a444044b/tasks/eva_2861/remediate_duplicate_dbsnp_ss/src/main/groovy/uk/ac/ebi/eva/remediate_duplicate_dbsnp_ss/RemediateDuplicateDbsnpSs.groovy#L51)) will fetch [all SVE documents that share the same ID](https://github.com/sundarvenkata-EBI/eva-accession/blob/c5ebaab9db6cbf48e1d104f0e60593ba7a4787f7/eva-accession-clustering/src/main/java/uk/ac/ebi/eva/accession/clustering/batch/io/SSSplitWriter.java#L229) from PROD.

tcezard: I see. 
Does it makes sense that the writer queries the database rather than relying on what the reader provides ? 
It is convenient in this case but I can see this be expensive if we were to reuse it elsewhere.

sundarvenkata-EBI: The expense is going to be the same. The reader still has to supply a set of SVEs along with their duplicates (needs sorting by accession in the reader - sacrifices restartability by imposing order on a non-shard key) + restrict the size of the result returned it so that they fit within the batch size: both of which can complicate the reader (at least when programming this as a Spring Batch reader with cursor). I do agree about the writer being extra zealous here.

sundarvenkata-EBI: Also, do let me know if we intend to reuse this writer.. If so, I can create a ticket to investigate refactoring this writer to a reader-writer process...

sundarvenkata-EBI: Agreed offline with @tcezard that there is no realistic expectation for this writer to be reused. But for clarity, [a note has been added to the SSSplitWriter's write method](https://github.com/EBIvariation/eva-accession/pull/345/commits/20edb4bbd181add7d5c1156b0d3e6600cb800f86) to document that only one of SVEs that share duplicate IDs can be passed to the writer.

nitin-ebi (approved):

apriltuesday (approved):

PR #78, opened 2022-07-14
EVA-2905 filter out null SVEs

tcezard:
SVEs that are filtered out by the processor and returned as null and needs to be removed list of processed variants


sundarvenkata-EBI (approved):

apriltuesday (approved):
Thanks for this fix!

PR #85, opened 2022-09-07
EVA-2950 - Script to detect variant that needs splits 

tcezard:
Starting from the output of EVA-2850 detect RS locus that needs to be split and the one that needs to be normalised.


tasks/eva_2950/split_rs_with_inconsistent_ss.py
```diff
@@ -0,0 +1,307 @@
...
+        if isinstance(node, (ast.Module, ast.Expr, ast.Dict, ast.Str, ast.List, ast.Constant,
+                            ast.Attribute, ast.Num, ast.Name, ast.Load, ast.Tuple)):
+            continue
+        if isinstance(node, ast.Call) and isinstance(node.func, ast.Attribute) and node.func.attr == 'datetime':
+            continue
+        raise ValueError(astr)
+    return eval(astr)
+
+
+def parse_eva2850_diagnositc_log(log_file):
```
sundarvenkata-EBI, line 171:
```suggestion
def parse_eva2850_diagnostic_log(log_file):
```

sundarvenkata-EBI (approved):

tasks/eva_2950/split_rs_with_inconsistent_ss.py
```diff
@@ -0,0 +1,307 @@
...
+    if len(ref) > 1 and len(alt) > 1:
+        return 'MNP'
+    if len(ref) > 1:
+        return 'DEL'
+    return 'INS'
+
+
+def shelve_submitted_variant_entities(mongo_handle, submitted_variant_ids):
+    output_collection = 'eva2950_dbsnpSubmittedVariantEntity'
+    query_match = {'_id': {'$in': submitted_variant_ids}}
```
sundarvenkata-EBI, line 216: Do we know how many ids are here? if there are a lot, the query below may fail...

tcezard: There are ~ 100K but might be more so I change the shelving method to do it in batch of 1000.

sundarvenkata-EBI (approved):

tasks/eva_2950/split_rs_with_inconsistent_ss.py
```diff
@@ -0,0 +1,320 @@
...
+def shelve_submitted_variant_entities(mongo_handle, submitted_variant_ids):
+    output_collection = 'eva2950_dbsnpSubmittedVariantEntity'
+    batch_size = 1000
+    for batch_sve_ids in grouper(submitted_variant_ids, batch_size):
+        query_filter = {'_id': {'$in': batch_sve_ids}}
+        documents = [d for d in mongo_handle["eva_accession_sharded"]['dbsnpSubmittedVariantEntity'].find(query_filter)]
+        mongo_handle["eva_accession_sharded"][output_collection].insert(documents)
+        nb_sve = mongo_handle["eva_accession_sharded"][output_collection].estimated_document_count()
+        assert nb_sve == len(batch_sve_ids), 'Not all variants were transfer to the output collection ' + output_collection
+        response = mongo_handle['dbsnpSubmittedVariantEntity'].delete_many(query_filter)
```
sundarvenkata-EBI, line 230: As we discussed some days back, we can just shelve the SS for now and do a delete after all the other shelving has been carried out in EVA-3004.

PR #99, opened 2022-12-14
EVA-3091 - New script to identify the assemblies where cross species clustering occurred

tcezard:


tasks/eva_3091/detect_cross_species_remapping.py
```diff
@@ -0,0 +1,138 @@
...
+                continue
+
+            # Check that if a clustering was ever performed on this assembly
+            if assembly not in assemblies_to_full_clustering_dates and \
+                    (taxonomy, assembly) not in assemblies_to_species_clustering_dates:
+                continue
+            dates_for_orig_assembly = ','.join([d.date().isoformat() if d else 'Unknown' for d in assemblies_to_full_clustering_dates.get(assembly, [])])
+            dates_for_orig_assembly += ','.join([d.date().isoformat() if d else 'Unknown' for d in assemblies_to_species_clustering_dates.get((taxonomy, assembly), [])])
+
+            # Check that if a clustering was ever performed on the other assembly
```
apriltuesday, line 113: Just to check my understanding: if we have `asm_1a` associated with `tax_1` but also containing variants from `tax_2`, and `tax_1`'s current supported assembly is `asm_1b`, then `asm_1a` is what we're calling the "original assembly" and `asm_1b` is the "other assembly".  Is that right?

tcezard: Yes that's correct, I should have a paragraph on nomenclature somewhere in the script/ticket

apriltuesday (approved):

sundarvenkata-EBI (approved):

sundarvenkata-EBI (approved):

PR #105, opened 2023-01-14
EVA-3117 - Check if non map-weight SS sharing ID with map-weight SS share RS loci also

sundarvenkata-EBI:


tasks/eva_3117/src/main/groovy/eva3117/eva3117_analyze_mapwt_ss_locus_sharing.groovy
```diff
@@ -0,0 +1,62 @@
...
+        svesWithoutMapWt.each{sveWithoutMapWt ->
+            if (rsHashesOfSvesWithMapWt.contains(getClusteredVariantHashSuppressErrors(sveWithoutMapWt))) {
+                logger.error("Non map-weight SS with hash ${sveWithoutMapWt.hashedMessage} and " +
+                        "accession ${sveWithoutMapWt.accession} shares ID and RS locus with a map-weighted SS")
+            }
+        }
+    }
+    numEntriesScanned += svesWithMapWt.size()
+    logger.info("Scanned ${numEntriesScanned} entries in ${assembly} so far....")
+}}
```
tcezard, line 62: I'm wondering if there are other way of writing/formatting this code in Groovy. I found it particularly difficult to follow the logic despite the small amount of code. 
For example could we format queries like:
```groovy
prodEnv.mongoTemplate.find(
    query(
        where("seq").is(assembly)
        .and("accession").in(svesWithMapWt.collect{it.accession})
        .and("mapWeight").exists(false)
    ), dbsnpSveClass
).groupBy {it.accession}
``` 
And extract the `def` as functions with parameters ?

tcezard (approved):

tasks/eva_3117/src/main/groovy/eva3117/eva3117_analyze_mapwt_ss_locus_sharing.groovy
sundarvenkata-EBI: I am trying to update the [Java coding style](https://www.ebi.ac.uk/seqdb/confluence/display/VAR/Java+coding+style) customizations to support Groovy code. I will keep you posted if I make any progress.

nitin-ebi (approved):

PR #113, opened 2023-02-08
EVA-3158 - Docker file wrapping around eva-submission validation

tcezard:
This docker file show what is needed to run submission automation.

### DIFF
--- a/tasks/eva_3158/Dockerfile
+++ b/tasks/eva_3158/Dockerfile
@@ -1,0 +1,58 @@
+FROM python:3.10
+
+ENV vcf_validator_version=0.9.4
+ENV python_version=3.10.9
+ENV submission_automation_version=v1.9.0
+ENV samtools_version=1.16.1
+ENV bcfools_version=1.16
+ENV htslib_version=1.16
+ENV NXF_VER=22.10.6
+
+WORKDIR /opt
+
+RUN mkdir vcf_validator \
+    && cd vcf_validator  \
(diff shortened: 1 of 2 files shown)


tasks/eva_3158/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+# Install python
+#RUN curl -LJO https://www.python.org/ftp/python/${python_version}/Python-${python_version}.tgz
+#RUN tar -zxf Python-${python_version}.tgz
+#RUN rm Python-${python_version}.tgz
+#RUN cd Python-${python_version}  \
+#    && ./configure --prefix=$(pwd) > configure.log
+#RUN cd Python-3.10.9 && make > make.log 2>&1
+#RUN cd Python-3.10.9 && make altinstall > install.log 2>&1
+#RUN cd ..
+#RUN Python-${python_version}/python -m venv ./venv
```
apriltuesday, line 29: I guess you don't need these as you're using a python image - maybe you're just keeping them to reference in the future?

apriltuesday (approved):

nitin-ebi (approved):

tasks/eva_3158/Dockerfile
tcezard: Yes that's right. I attempted to use a bare bone image but it turned out to be too time consuming so reverted to the fully fledge image to get something to work. 
Just leaving this here for future reference.

PR #136, opened 2023-06-26
EVA-3283 copy log files from 53_clustering/logs to 00_logs

nitin-ebi:


sundarvenkata-EBI (approved):

tasks/eva_3283/move_logs.py
```diff
@@ -0,0 +1,55 @@
...
+
+def main():
+    argparse = ArgumentParser(description='COPY logs from clustering to logs directory')
+    argparse.add_argument('--projects_dir_path', required=True, type=str, help='The Path to projects dir')
+    argparse.add_argument('--project', required=False, type=str,
+                          help='Project for which logs needs to be copied/deleted')
+    argparse.add_argument('--copy_logs', action='store_true', default=False,
+                          help='Copy logs from clustering directory to logs directory')
+    argparse.add_argument('--delete_clustering_logs', action='store_true', default=False,
+                          help='Delete logs from clustering directory')
```
tcezard, line 21: That will work, but I can't see the rational of running `move_logs.py` script without actually copying them. 
Also you take the risk of deleting the logs before they are copied.

tasks/eva_3283/move_logs.py
```diff
@@ -0,0 +1,55 @@
...
+        logger.info(f'Processing project {project}')
+        project_clustering_logs_dir = os.path.join(args.projects_dir_path, project, '53_clustering', 'logs')
+
+        if os.path.exists(project_clustering_logs_dir):
+            if args.copy_logs:
+                project_logs_dir = os.path.join(args.projects_dir_path, project, '00_logs')
+                if os.listdir(project_clustering_logs_dir):
+                    logger.info(f'Copying Files for project {project}')
+
+                    copy_command = f'cp {project_clustering_logs_dir}/* {project_logs_dir}'
```
tcezard, line 43: This is copying everything including the properties files that should stay in `53_clustering`

tasks/eva_3283/move_logs.py
```diff
@@ -0,0 +1,55 @@
...
+                    command_utils.run_command_with_output('copying log files', copy_command)
+                else:
+                    logger.warning(f'Logs directory is empty for project {project}')
+
+            if args.delete_clustering_logs:
+                shutil.rmtree(project_clustering_logs_dir)
```
tcezard, line 49: We should not delete the whole directory only the logs.

tcezard (changes requested):

tasks/eva_3283/move_logs.py
nitin-ebi: Properties files are directly under **53_clustering** while logs are in  **53_clustering/logs**. I am copying everything from **53_clustering/logs** which should not have anything apart from logs

tasks/eva_3283/move_logs.py
nitin-ebi: The logs are under a separate directory (**53_clustering/logs**), I am deleting this logs directory under **53_clustering** and not **53_clustering** itself

tcezard (approved):

PR #157, opened 2024-01-11
EVA-3450 - Create SPLIT and MERGE candidates in DEV from information from PROD

tcezard:


tasks/eva_3450/src/main/groovy/eva3450/CreateSplitCandidates.groovy
```diff
@@ -0,0 +1,121 @@
...
+
+    def createBatchSplitCandidates = { List<String> rsidAndHashes ->
+        Map<Long, String[]> rsidToHashes = rsidAndHashes.collectEntries {
+            String[] sp_line = it.split("\\t")
+            def rsid = sp_line.first().toLong()
+            def hashes = sp_line.tail()
+            return [ (rsid): hashes]
+        }
+        def rsids = rsidToHashes.keySet().toList()
+        def evaAndDbsnpSveCursors = [sveClass, dbsnpSveClass].collect { collectionClass ->
```
apriltuesday, line 77: You can use `collectMany` to simplify this a bit - example courtesy of Nitin [here](https://github.com/EBIvariation/eva-tasks/blob/1b8abeea8e626ca248aedbee3438e1f7de3afe8d/tasks/eva_3417/src/main/groovy/eva3417/RemediateLowerCaseRefAlt.groovy#L87)

tasks/eva_3450/src/main/groovy/eva3450/eva3450_create_merge_candidates.groovy
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.context.ConfigurableApplicationContext
+import uk.ac.ebi.eva.accession.core.GenericApplication
+import static uk.ac.ebi.eva.groovy.commons.EVADatabaseEnvironment.*
+
+
+// This script collects pending splits and merges from the given assembly in PROD.
+// If DEV environment is provided, will store them in
+// eva3372_accession_sharded.rsIdsWithCveHash and cveHashesWithRsId collections.
+// Otherwise will output intermediate results to file.
```
apriltuesday, line 14: might want to update the comments...

tasks/eva_3450/src/main/groovy/eva3450/CreateMergeCandidates.groovy
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    def createBatchMergeCandidates = { List<String> hashAndRsids ->
+        Map<String, Long[]> hashToRsids = hashAndRsids.collectEntries {
+            String[] sp_line =  it.split("\\t")
+            def hash = sp_line.first()
+            def rsids = sp_line.tail().collect({it.toLong()})
+            return [ (hash): rsids]
+        }
+        def rsids = hashToRsids.values().toList().flatten()
```
apriltuesday, line 69: `flatten` is not needed, and actually I don't think it needs to be converted to a list either - I think the query just needs a `Collection` which `values()` will give you.

apriltuesday (approved):
Great work, left some comments but they're optional or just meant as an FYI.

tasks/eva_3450/nextflow/run_batch.nf
```diff
@@ -0,0 +1,112 @@
...
+
+    script:
+    groovy_project_dir=file(params.load_merge_groovy).getParent().getParent().getParent().getParent().getParent()
+    """
+    run_groovy_script.sh $groovy_project_dir $params.load_merge_groovy -assemblyAccession=$params.assembly_accession -devPropertiesFile=$params.write_properties_file  -prodPropertiesFile=$params.read_properties_file -rsMergeCandidateFile=${source_duplicate_file.toRealPath()} > ${source_duplicate_file}_merged_candidate.log 2>&1
+    """
+}
+
+process load_split_candidate_variants {
+    memory '16 GB'
```
tcezard, line 68:
```suggestion
    memory '8 GB'
```

tasks/eva_3450/src/main/groovy/eva3450/CreateMergeCandidates.groovy
tcezard: I might be wrong but `hashToRsids.values()` give a `Collection` of `Long[]`. To get to a list of `Long`, I need to flatten it and I had to convert it to a List because `flatten()` didn't exist on `Collection`

apriltuesday: You're right, I forgot there was the extra `Long[]` level

nitin-ebi (approved):

PR #162, opened 2024-02-09
Update web counts

tcezard:
- Split counts between species/general download
- Add analysis per month

View the notebook [here](https://nbviewer.org/github/EBIvariation/eva-tasks/blob/ac2dd319fdd74c0f15f29f16e5425ed03a15604f/tasks/eva_2766/eva_data_usage_analysis_.ipynb)


apriltuesday (approved):
Looks good, I think you're right that pretty much all we can read into the odd FTP charts is that someone used it in 2020, but besides that usage has been low and essentially flat (I guess just Ensembl maybe?)

Also small typo in some of the y-axis labels, `Go` instead of `Gb`:

image

nitin-ebi (approved):

Dona094 (approved):

PR #175, opened 2024-08-20
EVA-2038  Run Import accession job for studies

nitin-ebi:


tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+            join taxonomy t on t.taxonomy_id = pt.taxonomy_id 
+            join project_analysis pa on pa.project_accession = pt.project_accession 
+            join analysis a on pa.analysis_accession = a.analysis_accession 
+            join assembly asm on asm.assembly_accession = a.vcf_reference_accession 
+            join analysis_file af on af.analysis_accession = pa.analysis_accession 
+            join file f on f.file_id = af.file_id 
+            where pt.project_accession = '{project_id}' 
+            and f.filename = '{os.path.basename(accession_report_file).replace(".accessioned.vcf", ".vcf")}'
+            and f.file_type IN ('vcf', 'VCF', 'vcf_aggregate', 'VCF_AGGREGATE')
+            """
```
apriltuesday, line 63: This query is probably fine, but I think you can save some joins by using the `assembly_set` table - see similar query [here](https://github.com/EBIvariation/eva-submission/blob/61f4da73214e871a4a98cbe4f2ce5c164d09debe/eva_submission/eload_status.py#L202) for checking eload status.

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+        else:
+            return None
+
+
+def run_import_accession_job_for_project(working_dir, params, project_id):
+    logger.info(f"Starting processing project: {project_id}")
+
+    project_path = os.path.join(params['project_dir'], project_id)
+    if not os.path.exists(project_path):
+        logger.error(f"Could not find the Project {project_id}. Please make sure the project {project_path} exists")
```
apriltuesday, line 76: For older studies, I guess we'll use [this script](https://github.com/EBIvariation/eva-submission/blob/61f4da73214e871a4a98cbe4f2ce5c164d09debe/bin/retrieve_archives_from_lts.py) to fetch these from LTS?

I wonder if it's worth including this and the very first step (determining the project IDs) in the automation, rather than just the accession load as you've done here... perhaps it's easier to manage these manually though.

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='', add_help=False)
+    parser.add_argument("--working-dir", help="/path/to/dir where all the logs and other files will be stored",
+                        required="true")
+    parser.add_argument("--params-file", help="/path/to/params/file containing path of nextflow, java app etc.",
+                        required="true")
+    parser.add_argument("--project-list", help="Comma separated project list e.g. PRJEB123,PRJEB456",
+                        required=True, nargs='+')
```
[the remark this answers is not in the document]
nitin-ebi: corrected

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+    args = parser.parse_args()
+
+    if not os.path.exists(args.working_dir):
+        os.makedirs(args.working_dir, exist_ok=True)
```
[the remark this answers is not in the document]
nitin-ebi: updated

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+                run_nextflow(working_dir, params, project_id, accession_report_file, db_name)
+            else:
+                logger.error(
+                    f'Could not find DB name for project {project_id} and accession_report_file: {accession_report_file}')
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='', add_help=False)
+    parser.add_argument("--working-dir", help="/path/to/dir where all the logs and other files will be stored",
+                        required="true")
```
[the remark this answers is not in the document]
nitin-ebi: done

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+    parser.add_argument("--params-file", help="/path/to/params/file containing path of nextflow, java app etc.",
+                        required="true")
```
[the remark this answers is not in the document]
nitin-ebi: done

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,114 @@
...
+from ebi_eva_internal_pyutils.pg_utils import get_all_results_for_query
+
+logger = log_cfg.get_logger(__name__)
+log_cfg.add_stdout_handler()
+log_cfg.set_log_level(logging.INFO)
+
+
+def run_nextflow(working_dir, params, project_id, accession_report_file, db_name):
+    random_string = ''.join(random.choice(string.ascii_letters) for i in range(6))
+    workflow_name = f"import_accession_{project_id}_{random_string}"
```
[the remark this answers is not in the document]
nitin-ebi: Updated.
Since we might have to run multiple jobs for multiple accession report files present in the same project, this was just to give each one a separate work directory under Project.
You are right using the accession report name would be more useful. have updated it.

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
nitin-ebi: Keeping it as it is for the time being.

Checked the query and it might save one join to taxonomy (will need to join assembly anyway to get assembly_accession as assembly_set has assembly_name and not accession), but the bigger problem is that assembly_set_id is null for a lot of analyses (in thousands - especially for old ones I guess)  while vcf_reference accession is present for all except for a handful which gives me more confidence that it will give the result.

apriltuesday: Makes perfect sense, thanks for checking.

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
tcezard: I think it's worth adding something that looks in the FTP. This will have to be done from a datamover node so you might have to run this script from datamover or retrieve the accession via a hard coded srun command.

tcezard (approved):
Looks good.
Left an optional comment

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
```diff
@@ -0,0 +1,162 @@
...
+            with get_metadata_connection_handle("production_processing", params['private_settings_xml_file']) as pg_conn:
+                acc_file_db_name_csv = os.path.join(working_dir, project_id, 'acc_file_db_name.csv')
+                with open(acc_file_db_name_csv, 'w', newline='') as file:
+                    writer = csv.writer(file)
+                    writer.writerow(['acc_report_file', 'db_name'])
+                    for acc_file, asm, taxonomy in file_asm_tax_list:
+                        db_name = resolve_variant_warehouse_db_name(pg_conn, asm, taxonomy)
+                        writer.writerow([formatted_name_file_dict[acc_file], db_name])
+
+                run_nextflow(working_dir, params, project_id, acc_file_db_name_csv)
```
apriltuesday, line 143:
```suggestion
            run_nextflow(working_dir, params, project_id, acc_file_db_name_csv)
```
since the db connection isn't needed anymore

apriltuesday (approved):

tasks/eva_2038/storeSSIDsInVariantWarehouse.py
nitin-ebi: done

PR #180, opened 2024-11-01
EVA-3660: Script to detect and update variants in variant warehouse that need normalisation

apriltuesday:


apriltuesday:
Tests are done, will do an end-to-end test in dev but meanwhile feel free to look at the code.

PR #181, opened 2024-12-04
EVA-3711 - Create duplicates

tcezard:
monitor_duplicate_accessions.py is a copy from [eva-tools](https://github.com/EBIvariation/eva-tools/blob/master/accession-monitoring/monitor_duplicate_accessions.py) with a new option to specify the study to extract
delete_variants_from_study.py remove all the ssids from the DEVA db 
duplicate_in_accession.nf automate the creation/assessment of the duplication

### DIFF
--- a/tasks/eva_3711/duplicate_in_accession.nf
+++ b/tasks/eva_3711/duplicate_in_accession.nf
@@ -4,29 +4,32 @@ nextflow.enable.dsl=2
 
 def helpMessage() {
     log.info"""
-    Process a chunk of merge or split candidates
+    Create duplicate during accessioning then detect them
 
     Inputs:
-            --source_duplicate_file     File containing list of duplicate (split or merge)
-            --clustering_props          Properties file for the Merge resolution (Clustering CLUSTER_UNCLUSTERED_VARIANTS_JOB)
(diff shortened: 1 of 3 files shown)


khetherin (approved):

tasks/eva_3711/duplicate_in_accession.nf
```diff
@@ -0,0 +1,103 @@
...
+
+    input:
+    val all_vcf_files
+
+    output:
+    val 1, emit: detect_duplicates
+
+    script:
+    """
+    $params.python $params.monitor_duplicate_accessions -p $params.accession_properties -s $params.study -o $params.output_dir -e eva-dev@ebi.ac.uk submittedVariantEntity
```
apriltuesday, line 101: Not sure we need it to email eva-dev, that email already gets bombarded with enough notifications that there's a risk of us ignoring it if a real alert comes through... I'd either make the email a parameter so you can set it to your own, or maybe even just change the message being sent so it's different from the prod alert.

apriltuesday (approved):

tasks/eva_3711/duplicate_in_accession.nf
tcezard: Done

PR #188, opened 2025-02-28
EVA-3754 - Script to categorise the RS duplicate candidates

tcezard:


nitin-ebi (approved):

tasks/eva_3754/categorise_duplicate_candidates.py
```diff
@@ -0,0 +1,88 @@
...
+def min_time_gap_between_create_date(cve_list):
+    date_list = [convert_date(cve.get('createdDate')) for cve in cve_list]
+    delta_dates = [abs(d1-d2) for d1,d2 in  itertools.combinations(date_list, 2)]
+    min_delta_date = min(delta_dates)
+    return min_delta_date
+
+def more_than_one_assembly(cve_list):
+    assembly_count = Counter([cve.get('assemblyAccession') for cve in cve_list])
+
+    return ','.join(sorted([assembly for assembly in assembly_count if assembly_count[assembly] > 1]))
```
apriltuesday, line 25: This is outputting something different from "Are all of these CVEs in one assembly or not", and it's a bit harder for me to interpret - e.g. cases where there are multiple assemblies, some of which might contain multiple CVEs and some might not. It doesn't matter in the end if the root causes/explanations are similar, but just something to point out...

apriltuesday (approved):

tasks/eva_3754/categorise_duplicate_candidates.py
tcezard: That's true the name of the function is wrong and should say something like. assemblies with more than one CVE.

PR #190, opened 2025-03-13
EVA-3764 Delete duplicate RS

nitin-ebi:


apriltuesday (approved):

tasks/eva_3764/delete_duplicate_rs_ids.py
```diff
@@ -0,0 +1,78 @@
...
+                f"Updated {update_result.modified_count} documents in submittedVariantEntity (removed rs field).")
+
+            # remove backpropagated RS in SVE
+            update_result = sve_coll.update_many(
+                {"backPropRS": {"$in": chunk}, "seq": "GCA_000188115.2", "study": "PRJEB82793"},
+                {"$unset": {"backPropRS": ""}}
+            )
+            logger.info(
+                f"Updated {update_result.modified_count} documents in submittedVariantEntity (removed backPropRS field).")
+
```
tcezard, line 47: I don't think this is the case but in theory these `rs` could have been used by other studies as well. 
Can we check if they are being used anywhere else in these two assemblies ?  
Maybe you can just add a `find` here and If we detect another submitted variant, we should run the same script for these other studies.

tcezard: That's fine then.

tcezard (approved):

PR #192, opened 2025-04-03
EVA-3779 - New script to delete variants that were remapped

tcezard:
Scripts to delete variants and load them back in the accessioning warehouse

### DIFF
--- a/tasks/eva_3779/ingest_remapped_vcf.nf
+++ b/tasks/eva_3779/ingest_remapped_vcf.nf
@@ -1,0 +1,48 @@
+
+
+
+
+/*
+ * Ingest the remapped submitted variants from a VCF file into the accessioning warehouse.
+ */
+process ingest_vcf_into_mongo {
+    label 'long_time', 'med_mem'
+
+    input:
+    each path(remapped_vcf)
+
+
+    output:
+    path "${remapped_vcf}_ingestion.log", emit: ingestion_log_filename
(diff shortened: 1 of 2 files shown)


tasks/eva_3779/ingest_remapped_vcf.nf
```diff
@@ -0,0 +1,48 @@
+
```
apriltuesday, line 1: Should we run the delete script as part of this nextflow, or will we just run it separately?

tasks/eva_3779/ingest_remapped_vcf.nf
```diff
@@ -0,0 +1,48 @@
...
+        > ${remapped_vcf}_ingestion.log
+    """
+}
+
+
+
+workflow {
+    main:
+        // Add list of vcf to load
+        remapped_vcfs = channel.fromPath( '' )
```
apriltuesday, line 44: Could also be passed in params, though for this script I think this is fine.

apriltuesday (approved):

nitin-ebi (approved):

tasks/eva_3779/ingest_remapped_vcf.nf
tcezard: I would prefer not to because the deletion will likely take a long time and therefore will need supervision.

PR #193, opened 2025-04-07
EVA-2326 Update contig name to INSDC

nitin-ebi:


tcezard (approved):
Code looks ok. Just left one suggestion

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
```diff
@@ -0,0 +1,578 @@
...
+import java.util.regex.Pattern
+import java.util.stream.Collectors
+
+import static org.springframework.data.mongodb.core.query.Criteria.where
+
+def cli = new CliBuilder()
+cli.workingDir(args: 1, "Path to the working directory where processing files will be kept", required: true)
+cli.envPropertiesFile(args: 1, "Properties file with db details to use for update", required: true)
+cli.dbName(args: 1, "Database name that needs to be updated", required: true)
+cli.assemblyReportPath(args: 1, "Path to the assembly report", required: true)
```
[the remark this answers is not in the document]
nitin-ebi: Didn't know about this method. Added now

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
```diff
@@ -0,0 +1,596 @@
...
+            List<VariantDocument> processInNextLoop = new ArrayList<>()
+
+
+            // create a map of original variant id and its INSDC accession chromosome
+            Map<String, String> orgVariantInsdcChrMap = variantDocumentList.stream()
+                    .collect(Collectors.toMap(orgVariant -> orgVariant.getId(), orgVariant -> {
+                        String orgChromosome = orgVariant.getChromosome()
+                        StringBuilder reason = new StringBuilder()
+                        ContigSynonyms contigSynonyms = contigMapping.getContigSynonyms(orgChromosome)
+                        if (contigMapping.isGenbankReplacementPossible(orgChromosome, contigSynonyms, reason)) {
```
apriltuesday, line 170: We've been using a more lenient version of this method, see [here](https://github.com/EBIvariation/eva-tasks/blob/ae261b7539a94414b30242cb188d47b7ecfc9486/tasks/eva_3660/src/main/groovy/eva3660/ContigRenamingProcessor.groovy#L60).

It actually might be worth re-using that class entirely, as it includes at least one set of workarounds (contig names in DB not present in assembly report) that we needed for contig mapping last time.

nitin-ebi: Done

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
```diff
@@ -0,0 +1,593 @@
...
+
+            // create a map of original variant id and its INSDC accession chromosome
+            Map<String, String> orgVariantInsdcChrMap = variantDocumentList.stream()
+                    .collect(Collectors.toMap(orgVariant -> orgVariant.getId(), orgVariant -> {
+                        String orgChromosome = orgVariant.getChromosome()
+                        try {
+                            return contigRenamer.getInsdcAccession(orgChromosome)
+                        } catch (ContigNotFoundException e) {
+                            // Log and bypass these errors
+                            logger.error("Could not get INSDC accession for variant {} with chromosome. Reason: {}", orgVariant.getId(),
```
apriltuesday, line 170:
```suggestion
                            logger.error("Could not get INSDC accession for variant {} with chromosome {}. Reason: {}", orgVariant.getId(),
```

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
```diff
@@ -0,0 +1,593 @@
...
+            logger.error("error storing variant id that can't be processed in the file:  {}", variantId)
+        }
+    }
+
+    VariantDocument remediateCaseNoIdCollision(VariantDocument orgVariant, String newId, String updatedChromosome) {
+        Map<String, Set<String>> updatedHgvs = getUpdatedHgvs(orgVariant, updatedChromosome)
+        Set<VariantSourceEntryMongo> updatedSourceEntries = orgVariant.getVariantSources().stream()
+                .peek(vse -> {
+                    if (vse.getAttrs() != null) {
+                        vse.getAttrs().append("CHR", updatedChromosome)
```
apriltuesday, line 344: Probably some discussion I forgot about, but would it make more sense to store the original contig name in the `files` attributes? Since the updated contig is already in the `chr` of the variant document.

Additionally, does appending to `attrs` in this way mean that it doesn't get included in the `src` attribute via [this method](https://github.com/EBIvariation/eva-pipeline/blob/master/src/main/java/uk/ac/ebi/eva/commons/models/mongo/entity/subdocuments/VariantSourceEntryMongo.java#L164)?  I'm not sure we'd want it to be included since it wasn't part of the original VCF, but just wanted to confirm that's what we want.

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
apriltuesday: Also, I don't think `reason` has been assigned yet...

apriltuesday (commented):
Looks great, just a couple things

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
nitin-ebi: corrected

tasks/eva_2326/src/main/groovy/eva2326/update_contig_name.groovy
nitin-ebi: removed

apriltuesday (approved):

PR #206, opened 2026-03-25
EVA-3933 Add script to read variant ids from a file and do the annotation remediatioin

nitin-ebi:


tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
```diff
@@ -0,0 +1,259 @@
...
+    void remediateAnnotations(Map<String, String> orgIdNewIdMap, Map<String, String> orgIdInsdcChrMap) {
+        Set<String> idProcessed = new HashSet<>()
+
+        // Build a combined regex query to fetch all relevant annotations in one round trip,
+        // matching documents whose _id starts with either the old or new variant id
+        List<Criteria> regexCriteria = new ArrayList<>()
+        for (Map.Entry<String, String> entry : orgIdNewIdMap.entrySet()) {
+            regexCriteria.add(Criteria.where("_id").regex("^" + Pattern.quote(entry.getKey()) + "_\\d"))
+            regexCriteria.add(Criteria.where("_id").regex("^" + Pattern.quote(entry.getValue()) + "_\\d"))
+        }
```
tcezard: Why not make two query per batch so that you know which one is old and which one is new?

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
```diff
@@ -0,0 +1,259 @@
...
+        Query combinedQuery = new Query(new Criteria().orOperator(regexCriteria.toArray(new Criteria[0])))
+
+        try {
+            List<Document> annotationsList = mongoTemplate.getCollection(ANNOTATIONS_COLLECTION)
+                    .find(combinedQuery.getQueryObject())
+                    .into(new ArrayList<>())
+
+            // Group fetched annotation documents by variant id prefix (both old and new ids)
+            Map<String, Set<Document>> variantIdToDocuments = new HashMap<>()
+            for (Document doc : annotationsList) {
```
tcezard: I might missunderstand this but it seems that for each annotation you go through all the document which would be 1000 * 1000 operations 
Wouldn't it be simpler to derive the `oldVariantId` and `newVariantId` from the `id` by removing the end and match with that?

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    Map<String, VariantDocument> getVariantsByIds(List<String> ids) {
+        Query idQuery = new Query(where("_id").in(ids))
+        List<VariantDocument> variants = mongoTemplate.find(idQuery, VariantDocument.class, VARIANTS_COLLECTION)
+        return variants.stream().collect(Collectors.toMap(v -> v.getId(), v -> v))
+    }
+
+    void storeNotRemediatedVariant(String oldVariantId, String newVariantId, String reason) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(notRemediatedVariantsFilePath, true))) {
```
tcezard: Do we need to open the file every time? If we have a lot of failure it could take a while.

apriltuesday (approved):

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
nitin-ebi: updated

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
nitin-ebi: updated

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
nitin-ebi: updated

tcezard (approved):

PR #208, opened 2026-04-02
EVA-3933 Add vep-version suffix to generate annotation ids

nitin-ebi:


tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
```diff
@@ -45,16 +44,30 @@ class RemediateAnnotationsApplication implements CommandLineRunner {
     private static int BATCH_SIZE = 1000
     public static final String VARIANTS_COLLECTION = "variants_2_0"
     public static final String ANNOTATIONS_COLLECTION = "annotations_2_0"
+    public static final Map<String, Set<String>> DB_VEP_CACHE_VERSION_SUFFIXES = [
+            "eva_hsapiens_grch37": ["78_78"].toSet(),
+            "eva_hsapiens_grch38": ["86_86"].toSet()
```
apriltuesday, line 49: I assume when we run this for the rest of the databases we'll need to query the annotation metadata collection for this? Unless there aren't that many, in which case hardcoding it is fine.

apriltuesday (approved):

tasks/eva_2326/src/main/groovy/eva2326/remediate_annotations.groovy
nitin-ebi: Yes, we'll need to query if there are too many, but I want to test this first on grch_37, and it has only one, so I have hardcoded for now.

tcezard (approved):

PR #209, opened 2026-04-27
EVA-4135 update annotations where chromosome in Id and chr field does not match

nitin-ebi:
There are 2 methods we can use, based on whether we want full control with fine-grained logging, or speed. 

We can do it in one shot using an aggregation pipeline, or we can go through each document and check the mismatched ones, modifying them and logging everything along the way.

We have to go through each document in both cases, but in aggregation, since we are doing everything inside the MongoDB, and there is no I/O over the network, it will be faster.

### DIFF
--- a/tasks/eva_4135/sync_chr_and_id_field.py
+++ b/tasks/eva_4135/sync_chr_and_id_field.py
@@ -1,0 +1,80 @@
+import argparse
+from ebi_eva_internal_pyutils.mongo_utils import get_mongo_connection_handle
+from ebi_eva_common_pyutils.logger import logging_config
+from pymongo import UpdateOne
+
+logging_config.add_stdout_handler()
+logger = logging_config.get_logger(__name__)
+
+BATCH_SIZE = 1000
+ANNOTATION_COLL_NAME = 'annotations_2_0'
+
+
+def process_annotations(db_name, annotation_coll):
+    modified_count = 0
+    cursor = annotation_coll.find({}, {'_id': 1, 'chr': 1}).batch_size(BATCH_SIZE).allow_disk_use(True)
+
+    batch = []
+    for variant in cursor:
+        batch.append(variant)
+        if len(batch) < BATCH_SIZE:
+            continue
+
+        # process variants in batch
+        modified_count += process_batch(db_name, batch, annotation_coll)
+        batch = []
+
+    # process remaining variants
+    if batch:
+        modified_count += process_batch(db_name, batch, annotation_coll)
+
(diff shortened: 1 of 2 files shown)


tasks/eva_4135/sync_chr_and_id_field.py
```diff
@@ -0,0 +1,80 @@
...
+
+def process_batch(db_name, batch, annotation_coll):
+    ids_modified = []
+    bulk_ops = []
+    for variant in batch:
+        chr_from_id = variant.get('_id').split("_")[0]
+        chr_from_field = variant.get('chr')
+        if chr_from_id != chr_from_field:
+            bulk_ops.append(UpdateOne({'_id': variant['_id']}, {'$set': {'chr': chr_from_id}}))
+            ids_modified.append(variant['_id'])
```
tcezard, line 42: I guess in the case of the `_id` being wrong and the `chr` being correct we will have to run the other remediation to fix it.
Should we check if chr match an INSDC accession regex and log the one that do not ?

tcezard: [vcf-validator](https://github.com/EBIvariation/vcf-validator/blob/de03ec2b9c180ad6be5d0c1274cf4529a6303457/inc/assembly_report/assembly_report.hpp#L89) uses:
```
"[A-Z][0-9]{5}\.[0-9]+"
"[A-Z]{2}[0-9]{6}\.[0-9]+"
"[A-Z]{2}[0-9]{8}\.[0-9]+"
"[A-Z]{4}[0-9]{8}\.[0-9]+|[A-Z]{6}[0-9]{9}\.[0-9]+"
```

tasks/eva_4135/sync_chr_and_id_field_using_expr.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for db_name in db_list:
+            if db_name in ['admin', 'config', 'local']:
+                continue
+            db = mongo_conn[db_name]
+            if ANNOTATION_COLL_NAME in db.list_collection_names():
+                logger.info(f"Processing Database: {db_name}")
+                annotation_coll = db[ANNOTATION_COLL_NAME]
+                # update mis-matched annotations
+                result = annotation_coll.update_many(
```
tcezard, line 34: I must say I'm a bit worried about this approach. We've not had good results (performance wise) when we've been delegating all the processing to MongoDB.
This being said, it looks correct.

tcezard (approved):
Which one of these script to  you think we should run ?

PR #211, opened 2026-04-30
EVA-4135 Fix chromosome mismatch in _id and Chr field - also update chromosome to INSDC

nitin-ebi:


apriltuesday (approved):

tasks/eva_4135/src/main/groovy/eva4135/update_annotations.groovy
```diff
@@ -0,0 +1,327 @@
...
+                .filter(annotUpdate -> !variantsIdsInDB.contains(annotUpdate.getVariantId()))
+                .collect(Collectors.toList())
+
+        if (!annotationsToBeUpdatedWithInsdcChromosomes.isEmpty()) {
+            List<WriteModel<Document>> bulkUpdateOperations = new ArrayList<>()
+
+            for (AnnotationUpdateModel annotationUpdateModel : annotationsToBeUpdatedWithInsdcChromosomes) {
+                // update annotation id and chr to be INSDC
+                Document updatedAnnotDocument = new Document(annotationUpdateModel.getOrginalAnnotationDocument())
+                String newId = annotationUpdateModel.getAnnotationId().replace(annotationUpdateModel.getChromosome(), annotationUpdateModel.getInsdcChromosome())
```
tcezard: Isn't it a bit risky to use replace here ? Especially if the original chromosome name is just `1` or `2`. You could end up replacing in the position as well.

nitin-ebi: Updated to replace the first occurrence at the start of the string.

tcezard (approved):

PR #212, opened 2026-05-07
Script to extract and summarise API usages and query parameters used

tcezard:
summarise_endpoints.py: List the endpoints being used and how much they are used after removing variable parameters in the URLs
summarise_query_params.py: List the query parameters of a subsets of endpoints most likely to use annotations

### DIFF
--- a/tasks/eva_4001/summarise_endpoints.py
+++ b/tasks/eva_4001/summarise_endpoints.py
@@ -30,8 +30,14 @@ _RULE_KEYS = set(CONTEXT_RULES.keys())
 # All proper prefixes of rule keys — used to know when to keep accumulating segments.
 _RULE_KEY_PREFIXES = {k[:i] for k in _RULE_KEYS for i in range(1, len(k))}
 
+BATCH_SIZE = 10_000
 
 def normalise_path(path):
+    """
+    Normalise a path by splitting it into segments and applying rules to replace segments with placeholders.
+    E.g. /eva/webservices/contig-alias/v1/chromosomes/md5checksum/7b6e06758e53927330346e9e7cc00cce
+     ==> /eva/webservices/contig-alias/v1/chromosomes/md5checksum/{md5}
+    """
     segments = path.split('/')
     normalised = []
     pending_key = ()   # rule-key prefix accumulated so far
@@ -66,10 +72,6 @@ def normalise_path(path):
             pending_key = ()
 
     return '/'.join(normalised)
-
-
-
-BATCH_SIZE = 10_000
 
 
 def stream_rows(conn, query):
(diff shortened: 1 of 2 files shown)


tasks/eva_4001/summarise_endpoints.py
```diff
@@ -0,0 +1,147 @@
...
+                normalised.append(CONTEXT_RULES[pending_key])
+            else:
+                normalised.append(seg)
+            pending_key = ()
+
+    return '/'.join(normalised)
+
+
+
+BATCH_SIZE = 10_000
```
apriltuesday: Would put this at the top with the other constants...

tasks/eva_4001/summarise_endpoints.py
apriltuesday, line 1: I think these scripts could be useful in the future, should we put them in tools instead of tasks? Maybe here: https://github.com/EBIvariation/eva-tools/tree/master/eva-usage-stats

tasks/eva_4001/summarise_endpoints.py
```diff
@@ -0,0 +1,147 @@
...
+    ('genbank',):                   '{genbank_accession}',
+    ('chromosomes', 'genbank'):     '{genbank_accession}',
+}
+
+_RULE_KEYS = set(CONTEXT_RULES.keys())
+# All proper prefixes of rule keys — used to know when to keep accumulating segments.
+_RULE_KEY_PREFIXES = {k[:i] for k in _RULE_KEYS for i in range(1, len(k))}
+
+
+def normalise_path(path):
```
apriltuesday, line 35: A docstring might be useful here - something like (if I understand the method correctly!):
```
Normalises paths by replacing path parameters with placeholders.
E.g. /eva/webservices/contig-alias/v1/chromosomes/md5checksum/7b6e06758e53927330346e9e7cc00cce --> /eva/webservices/contig-alias/v1/chromosomes/md5checksum/{md5}
```

tasks/eva_4001/summarise_query_params.py
```diff
@@ -0,0 +1,72 @@
...
+    print()
+    print(f"Fetched {total_rows:,} rows total.\n")
+    print(f"=== Query parameter usage since {args.date} ===\n")
+
+    for endpoint in sorted(TARGET_ENDPOINTS):
+        n = endpoint_counts[endpoint]
+        print(f"{endpoint}  ({n:,} requests)")
+        params = param_counts[endpoint]
+        if not params:
+            print("No query parameters observe)")
```
apriltuesday:
```suggestion
            print("No query parameters observed")
```

apriltuesday (commented):
LGTM, just wondering if these should go in eva-tools so it's easier to find and reuse them in the future?

nitin-ebi (approved):

apriltuesday (approved):

PR #214, opened 2026-05-18
EVA-4173 update script

nitin-ebi:


apriltuesday (approved):

tasks/eva_4135/src/main/groovy/eva4135/update_annotations.groovy
tcezard, line 163: We could retrieve the variant start and end and check that they are the same as the one present in the annotation.

tcezard, line 235: We should filter to keep the one with different starts as well.

tcezard (approved):

